PR #13, opened 2018-03-06
Add simple test for container + fix basics so that Jenkins can start

batmat:
Jenkins should be available after startup.

I have the changes locally that makes this work, though not clean (at all) yet:

```
evergreen $ ./test-container.sh
Using the port 5901
Start container under test and wait a bit for its startup:
b194fb14217a1174a43d888d70942328161419e21fdaaeee2688a9fc967a05d7
Connect to Jenkins
Check content
```

cc @rtyler

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,12 +7,13 @@ ARG gid=1000
 ARG http_port=8080
 ARG agent_port=50000
 
+ENV JAVA_OPTS -Djava.awt.headless=true
 ENV JENKINS_HOME /var/jenkins_home
 ENV JENKINS_AGENT_PORT ${agent_port}
 ENV EVERGREEN_ENDPOINT=http://127.0.0.1:9292
 ENV COPY_REFERENCE_FILE_LOG $JENKINS_HOME/copy_reference_file.log
+RUN mkdir -p /usr/share/jenkins/ref/init.groovy.d
 
-ENV JAVA_OPTS -Djava.awt.headless=true
 # Jenkins home directory is a volume, so configuration and build history
 # can be persisted and survive image upgrades
 VOLUME ${JENKINS_HOME}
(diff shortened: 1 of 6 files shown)


rtyler:
@batmat won't this fail every pull request until the evergreen-client is properly downloading the `jenkins.war` and starting it?

batmat:
@rtyler yes, I meant it as a first commit. It is expected to fail as is indeed. Going to see how to push a bare minimum to make it pass (probably with some `FIXME` hacks)

rtyler:
Heh, the Pipeline didn't fail, so perhaps the test is bad? :smile:

batmat:
THAT WAS THE INTENT, AND EXPECTED :+1: 

/me goes back fixing this crap. ;-)

batmat:
Better, **now** it fails as expected.
image

supervisord.conf
```diff
@@ -8,7 +8,7 @@ port=:9001
 supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
 
 [program:jenkins]
-command=/usr/local/bin/jenkins.sh
+command=/usr/local/bin/shim-startup-wrapper.sh
```
batmat: @rtyler WDYT of this as a temporary measure to be able to move forward? 

At least it's very useful for me locally to be able to iterate.

Dockerfile
```diff
@@ -53,6 +60,9 @@ RUN mkdir -p /evergreen
 COPY client /evergreen/client
 COPY essentials.yaml /evergreen
 
+# FIXME REMOVE (to ease iteration/speed just for now), see also shim-startup-wrapper.sh
+RUN wget --quiet http://mirrors.jenkins.io/war-stable/latest/jenkins.war -O /usr/share/jenkins/jenkins.war
```
batmat: complementary to the shim. The goal here is to try and leverage Docker cache and avoid having to download the WAR **each** time one would start the container. Which is already a few seconds on a high bandwidth network, but can kill everything on a more flaky network when hacking on evergreen.

PR #14, opened 2018-03-07
Relocate scripts into the scripts/ directory

rtyler:
This commit also cleans up the Makefile a bit to ensure that the container is
always built before we execute the container-check target.


scripts/test-container.sh
```diff
@@ -27,7 +27,6 @@ find_free_port() {
 }
 
 ### HERE STARTS THE REAL MEAT
-docker
```
batmat: wooops :-\

PR #17, opened 2018-03-10
Jotting down some basic thoughts about what evergreen-client is responsible for

rtyler:
This is obviously not complete, but after conversations with @batmat last week I
figure it's important to get _something_ of the thoughts in my notebook written down.

### DIFF
--- a/client/README.adoc
+++ b/client/README.adoc
@@ -1,0 +1,49 @@
+= evergreen-client
+
+This directory contains the `evergreen-client` application which provides the
+client-side functionality necessary for the Evergreen distribution system.
+
+The purpose of `evergreen-client` is to facilitate the following:
+
+* Maintain <<status>> with the Evergreen hosted services layer.
+* Orchestrate <<upgrades>> of the `jenkins.war` and plugin `.jpi` files when
+  instructed by the Evergreen hosted services layer.
+* Collect and deliver telemetry as necessary to the Evergreen hosted services
+  layer.
+* Additional capabilities as the needs by Jenkins Essentials evolve.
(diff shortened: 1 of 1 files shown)


client/README.adoc
```diff
@@ -0,0 +1,49 @@
...
+[[upgrades]]
+=== Upgrading
+
+NOTE: This is still a work in progress
+
+Types of upgrades `evergreen-client` should be responsible for:
+
+* An evergreen manifest (i.e. Jenkins core and plugins), including adding and
+  removing plugins as the needs of Jenkins Essentials changes
+* Configuration as Code to manage the "automatic sane defaults" aspect of
```
batmat: Intends to _as much as possible_ I guess. Maybe it's OK here as a design goal to just document somehow here that we _aim_ to use it as much as possible.

My point is: I'm sure we'll need to do things _ourselves_ at least while things are being improved/fixed upstream (with JIRA issues filed upstream I guess).

client/README.adoc
```diff
@@ -0,0 +1,49 @@
...
+  Jenkins Essentials.
+* `evergreen-client` releases itself.
```
batmat: Took me some time to get this. I was confused by _releases_. I think putting "`evergreen-client` itself." only would have avoided my confusion.

client/README.adoc
rtyler: We can evaluate those cases when they come up. I'm not going to give us a free ticket to implement hacks just because we're both _very_ comfortable implementing hacks in Jenkins :smiling_imp:

PR #19, opened 2018-03-14
New yaml format for essentials.yaml

rtyler:
There is still some tooling which needs to support this outside of the scope of this pull request, but I figured this would be a good start.


essentials.yaml
```diff
@@ -1,378 +1,63 @@
...
-    url: https://updates.jenkins.io/download/plugins/workflow-scm-step/2.6/workflow-scm-step.hpi
-  - name: workflow-step-api
-    version: '2.14'
-    gav: org.jenkins-ci.plugins.workflow:workflow-step-api:2.14
-    url: https://updates.jenkins.io/download/plugins/workflow-step-api/2.14/workflow-step-api.hpi
-  - name: workflow-support
-    version: '2.18'
-    gav: org.jenkins-ci.plugins.workflow:workflow-support:2.18
-    url: https://updates.jenkins.io/download/plugins/workflow-support/2.18/workflow-support.hpi
+metadata:
```
batmat: I'm probably missing a part of this, and it's definitely enough for now. But just for the future and my understanding: where would we say (would we?) that a given file is for the `canary` channel, `beta`, etc. ?

essentials.yaml
```diff
@@ -1,378 +1,63 @@
...
+  labels:
+    name: 'jenkins-essentials'
+  annotations:
```
batmat: can/should we add a comment or so to explain what those are designed to do?

Without much context, it's hard to know what difference you're making and what its goal is for `labels` vs. `annotations`?

essentials.yaml
```diff
@@ -1,378 +1,63 @@
...
+spec:
```
batmat: same with `spec` I guess

essentials.yaml
```diff
@@ -1,378 +1,63 @@
...
+    # version: 1.0
+
+  plugins:
+    - groupId: 'io.jenkins.blueocean'
+      artifactId: 'blueocean'
+      #ref: 'master'
+      version: '1.4.2'
+
+    - groupId: 'org.jenkins-ci.plugins'
+      artifactId; 'git'
```
batmat: typo, I guess you want `:`, not `;` :bug: 

I think we should add some step in the `Jenkinsfile` like `find . -name "*.yaml" -exec yamllint {} \;`

rtyler:
Woops! I'll update the `Jenkinsfile` and make some fixes :smile:

essentials.yaml
bitwiseman: @batmat - 
For things like `;` to `:`, you might consider editing directly.  

My theory is that for obvious typos it takes less combined time for the commenter to fix the issue directly in the GitHub "Edit" UI as it does to make the comment, have the submitter make  the fix, and then review again.

Thoughts?

rtyler:
@batmat generally speaking this is just copying some of the manners in which Kubernetes resources are defined. I don't think @carlossg nor myself will fully know what we're going to need until there's a bit more playing around and prototyping done with this.

Makefile
```diff
@@ -8,7 +8,10 @@ SCRIPTS_URL=https://raw.githubusercontent.com/jenkinsci/docker/master/
 #################
 all: check container
 
-check:
+lint: essentials.yaml
+	./tools/yamllint -s ./essentials.yaml
```
batmat: :+1:

essentials.yaml
batmat: Well, possibly yes for some specific cases, maybe here.
As a general rule I think I disagree: the submitter might realize something deeper being wrong, or to fix. And s/he may not fully like the fact someone else has inserted commits in the middle, that might have to be force pushed/rebased or whatever. (and as some people still sometimes get lost using git and prefer to open a new PR and close the previous, just because of conflicts/rebase/merge issue they don't really know how to handle :P).

And here, the right thing to do IMO, which Tyler did do, was also to add a test to catch this in the future, so was not _that_ much a small fix.

PR #24, opened 2018-03-16
Enable shellcheck for shell scripts

batmat:
A bit more code quality enforcement/check, for shell scripts this time using `shellcheck`.

### DIFF
--- a/tests/tests.sh
+++ b/tests/tests.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
+# Note: would have used set -euo pipefail, but ./shunit2 unfortunately fails hard with this :-(.
 
-. $(dirname $0)/utilities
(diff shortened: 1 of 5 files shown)


rtyler (approved):

tests/tests.sh
```diff
@@ -11,18 +16,18 @@ oneTimeTearDown() {
 }
 
 test_smoke() {
-  curl --silent http://localhost:$TEST_PORT > /dev/null
+  curl --silent "http://localhost:$TEST_PORT > /dev/null"
```
batmat: How the heck did you succeed in Jenkins?! It fails for me locally now with an exit code 3, i.e. URL malformed, which is what I would have expected then...

PR #25, opened 2018-03-18
Build and use master branch of CasC

batmat:
(Note: not sure that `mvn` and `git clone` raw command calls are going to work in CI. Let's see.)

CasC being still in very early phase, using `master` will help
us benefit immediately from improvements or fixes filed upstream (like https://github.com/jenkinsci/configuration-as-code-plugin/pull/152 for example).

This being too the model we target in the long term, this is not
probably not that weird to try it (though the delivery model will
obviously be very different)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,3 +84,5 @@ ENV CASC_JENKINS_CONFIG=$JENKINS_HOME/jenkins.yaml
 # that the Jenkins and evergreen-client processes both execute properly
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 CMD /usr/bin/supervisord -c /etc/supervisor/conf.d/supervisord.conf
+
+COPY configuration-as-code.hpi /usr/share/jenkins/ref/plugins/configuration-as-code.hpi
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,6 +77,7 @@ COPY essentials.yaml /evergreen
 
 # FIXME (?): what if the end users touches the config value?
 # as is, we'll override it.
+COPY build/configuration-as-code/target/configuration-as-code.hpi /usr/share/jenkins/ref/plugins/configuration-as-code.hpi
 COPY jenkins-configuration.yaml /usr/share/jenkins/ref/jenkins.yaml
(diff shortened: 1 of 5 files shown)


batmat:
OK, as I _feared_ I need to fix/evolve the `Jenkinsfile` to teach CI about how to find & use mvn.

rtyler:
Use the Docker, Luke

batmat:
if you mean `docker.inside`, never :P. But yes, maybe some form of it.

Dockerfile
```diff
@@ -84,3 +84,5 @@ ENV CASC_JENKINS_CONFIG=$JENKINS_HOME/jenkins.yaml
 # that the Jenkins and evergreen-client processes both execute properly
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 CMD /usr/bin/supervisord -c /etc/supervisor/conf.d/supervisord.conf
+
+COPY configuration-as-code.hpi /usr/share/jenkins/ref/plugins/configuration-as-code.hpi
```
batmat: :man_facepalming: wrong path now, test is going to fail.

Makefile
```diff
@@ -51,6 +53,12 @@ build/install-plugins.sh:
 	curl -sSL $(SCRIPTS_URL)/install-plugins.sh > $@
 	chmod +x $@
 
+build/configuration-as-code:
+	git clone https://github.com/jenkinsci/configuration-as-code-plugin.git build/configuration-as-code
+
+build/configuration-as-code/target/configuration-as-code.hpi: build/configuration-as-code
+	cd build/configuration-as-code && mvn clean package -DskipTests
```
orrc: If you're not planning to do anything special with the source, may as well just pull the .hpi from [J-o-J](https://ci.jenkins.io/job/Plugins/job/configuration-as-code-plugin/job/master/lastStableBuild/artifact/target/)?

rtyler (changes requested):
By "use Docker" I meant wrap a script in `./tools`. Right now all a new contributor needs is `docker` and `make` in order to build and test the whole project, which is worth keeping IMHO

batmat:
@rtyler fixed.

rtyler (approved):

PR #27, opened 2018-03-20
[JENKINS-49406] Start segregating data a bit more

batmat:
This is more prototyping than a final version as the JEP is still
ongoing (https://github.com/batmat/jep/pull/1). But this has value to show the JEP is actually doable.

To sum up, introduce `/var/jenkins` as new root, and have subdirs there as discussed on the list.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,19 +7,29 @@ ARG gid=1000
 ARG http_port=8080
 ARG agent_port=50000
 
-ENV JAVA_OPTS -Djava.awt.headless=true
-ENV JENKINS_HOME /var/jenkins_home
+# Note: new env var, specific to Essentials. Not strictly necessary, but will help avoid
+# hardcode this path everywhere
+# *NOT* /var/jenkins_home on purpose, because we probably want people to ack the structure is
+# not the same as the jenkins/jenkins image
+ENV JENKINS_DIR /var/jenkins
+ENV JENKINS_HOME ${JENKINS_DIR}/home
 ENV JENKINS_AGENT_PORT ${agent_port}
 ENV EVERGREEN_ENDPOINT=http://127.0.0.1:9292
 ENV COPY_REFERENCE_FILE_LOG $JENKINS_HOME/copy_reference_file.log
 ENV JENKINS_UC https://updates.jenkins.io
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -7,19 +7,29 @@ ARG gid=1000
...
+ENV JENKINS_DIR /var/jenkins
+ENV JENKINS_HOME ${JENKINS_DIR}/home
 ENV JENKINS_AGENT_PORT ${agent_port}
 ENV EVERGREEN_ENDPOINT=http://127.0.0.1:9292
 ENV COPY_REFERENCE_FILE_LOG $JENKINS_HOME/copy_reference_file.log
 ENV JENKINS_UC https://updates.jenkins.io
 ENV JENKINS_UC_EXPERIMENTAL=https://updates.jenkins.io/experimental
 
-RUN mkdir -p /usr/share/jenkins/ref/init.groovy.d
+ENV JAVA_OPTS="-Djava.awt.headless=true -Djenkins.model.Jenkins.WORKSPACES_DIR=${JENKINS_DIR}/var/\${ITEM_FULL_NAME}/workspace -Djenkins.model.Jenkins.BUILDS_DIR=$JENKINS_DIR/var/\${ITEM_FULL_NAME}/builds"
```
batmat: the `jenkins.model.Jenkins.*_DIR` won't do anything until https://github.com/jenkinsci/jenkins/pull/3360 or some form of it gets merged and we can use the tip of the core.

supervisord.conf
```diff
@@ -9,8 +9,7 @@ supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
 
 [program:jenkins]
 command=/usr/local/bin/shim-startup-wrapper.sh
-directory=/var/jenkins_home
-environment=JENKINS_HOME=/var/jenkins_home
```
batmat: Removed this ^ since supervisord is documented and does inherit those, so it's a duplicate of the JENKINS_HOME value we better avoid.

batmat:
Need to adjust this with latest changes in https://github.com/jenkinsci/jep/pull/67

Waiting a bit for the JEP to be reviewed and settle before moving forward here to avoid adjusting, then re-adjusting, then re-re-adjusting :yum:.

PR #35, opened 2018-03-28
Use latest tip of the master branch built in CI

batmat:
For now as we don't have a better place than Jenkins itself to download it from.


batmat:
\o/ build success

PR #37, opened 2018-03-29
Client registration

rtyler:
This pull request actually represents the research work I initially did pre-dating the design discussed in [this thread](https://groups.google.com/d/msgid/jenkinsci-dev/20180324020555.p26pfpln3j722m5r%40blackberry.coupleofllamas.com).

I've opted to leave much of the reverted works in progress in the history because I think it may be useful for future code archaeologists :smile:


rtyler:
Hrm, the acceptance tests for services are having trouble binding an address :frowning_face:

rtyler:
@batmat I've added you as a reviewer, don't feel obligated to force yourself to understand too much JavaScript here.

I think I'd like to call "registration" ready and get that merged before continuing on the "login" part of the flow

client/.eslintrc.json
```diff
@@ -0,0 +1,29 @@
+{
+  "env": {
+    "es6": true,
+    "node": true,
+    "jest": true
+  },
+  "parserOptions": {
+    "ecmaVersion": 2017
```
batmat: NIT: I don't know what I'm talking about, seems like from reading in https://eslint.org/docs/user-guide/configuring that 2018 does exist. 
Maybe a JS PR I could handle :P

client/client.js
```diff
@@ -1,55 +1,49 @@
...
-
-const sse = new EventSource(util.format('%s/sse/stream/%s', ENDPOINT, ident));
-console.debug('EventSource created', sse);
-
-/* First set up a generic event handler.
- *
- * This function will be called whenever an unnamed SSE is received, an that
- * should typically never happen.
+/*
+ * This is the main entryooint for the evergreen-client
```
batmat: typo

client/client.js
```diff
@@ -1,55 +1,49 @@
...
-/* Currently not exporting any symbols */
-module.exports = () => {
+    }
+    else {
+      logger.info('Registering..');
+      reg.register().then((res) => {
+        /* successfully created registration */
+      }).catch((err) => {
+        logger.error('Failed to complete a registration, what do we do!', err);
+      });;
```
batmat: double `;` ?

client/lib/registration.js
```diff
@@ -0,0 +1,257 @@
...
+
+    return true;
+  }
+
+  /*
+   * Load generated keys from disk assuming they are present
+   *
+   * @return Boolean on successful load of keys
+   */
+  loadKeysSync() {
```
batmat: Is this called from anywhere but tests?

client/lib/registration.js
```diff
@@ -0,0 +1,257 @@
...
+
+  /*
+   * Returns the default home directory or the value of EVERGREEN_HOME
+   */
+  homeDirectory() {
+    /* The default home directory is /evergreen, see the Dockerfile in the root
+     * directory of th repository
+     */
+    if (!process.env.EVERGREEN_HOME) {
+      return '/evergreen';
```
batmat: Shouldn't we rather throw an exception? Or do you use that I guess when developing? Maybe it should log a warning ?

client/package.json
```diff
@@ -5,19 +5,26 @@
...
+    "@feathersjs/feathers": "^3.1.4",
+    "@feathersjs/rest-client": "^1.3.4",
     "eventsource": "^1.0.5",
-    "node-fetch": "^2.0.0",
+    "mkdirp": "^0.5.1",
+    "node-fetch": "^2.1.2",
     "unzip": "^0.1.11",
     "uuid": "^3.2.1",
+    "winston": "^2.4.1",
     "xmlrpc": "^1.3.2"
```
batmat: IIUC, `^` means we trust all those dependencies do use semver correctly, and never screw up right? I'll ask friends in that field, but wouldn't we want fixed versions?

client/test/registration.test.js
```diff
@@ -0,0 +1,181 @@
...
+  describe('generateKeys()', () => {
+    it('should return a boolean on success', () => {
+      assert.ok((new Registration()).generateKeys());
+    });
+  });
+
+  describe('publicKeyPath()', () => {
+    it('should return a path', () => {
+      const p = (new Registration()).publicKeyPath();
+      assert(p != path.basename(p), 'This doesn\'t look like a path');
```
batmat: IIUC, the fact you cannot use `"This doesn't look like a path"` here is because of the esling setting done above at https://github.com/jenkins-infra/evergreen/pull/37/files#diff-d89f635f668d643667d79f8fcc5f9f41R20

client/test/registration.test.js
```diff
@@ -0,0 +1,181 @@
...
+    });
+  });
+
+  describe('keyPath()', () => {
+    it('should return a path', () => {
+      const keys = (new Registration()).keyPath();
+      assert(keys != path.basename(keys), 'This doesn\'t look like a path');
```
batmat: Should we typically pass the path to the assertion for diagnosability ? (IIUC ES 6 allows using backticks to do interpolation and make it bit more readable?)

client/package.json
rtyler: package-lock.json is where fixed versions are already locked.

client/test/registration.test.js
rtyler: Yeah, and that is okay :smile:

client/lib/registration.js
rtyler: This is not called yet because it's required for the Login flow, this code isn't quite there yet.

client/.eslintrc.json
rtyler: This was something set originally with the feathersjs generator, I don't know enough about how ECMAScript 2018 relates to functionality provided by Node 9. There are some things in newer versions of ECMAScript which are "Experimental" in Node 9, but that doesn't mean we should rely on them yet.

client/test/registration.test.js
rtyler: That's a good suggestion, I was vaguely aware of [Template strings](http://tc39wiki.calculist.org/es6/template-strings/) but haven't gotten myself in the habit of using them yet.

client/package.json
batmat: OK, yes indeed.

PR #42, opened 2018-04-03
Implementing post-registration client login

rtyler:
See: 
* [JENKINS-50488](https://issues.jenkins-ci.org/browse/JENKINS-50488)
* [JENKINS-50494](https://issues.jenkins-ci.org/browse/JENKINS-50494)

This pull request effectively implements the design proposed in [JEP-303](https://github.com/jenkinsci/jep/blob/master/jep/303)


rtyler:
These changes introduce a number of hooks as anonymous inline functions, this is technical debt which I am willing to accrue right now. They should be refactored out and unit tested in the future, since they're presently only covered by acceptance tests.

services/src/services/authentication/authentication.hooks.js
```diff
@@ -0,0 +1,93 @@
...
+      /*
+       * Using the public key passed by the previous hook, ensure that the
+       * signature of the UUID matches appropriately
+       */
+      (hook) => {
+        let record = hook.data.record;
+        let ec = new ecc.ec(record.curve);
+        let key = ec.keyFromPublic(record.pubKey, 'hex');
+
+        let invalidError = new errors.BadRequest('Invalid signature provided');
```
batmat: NIT (mostly for my understanding really): I don't measure well the impact of creating this each time, but IIUC this is going to be always run, when it's supposed to be useful/used when something goes wrong. I hope/suppose the call is not too costly. As it's the service side, I guess this is not going to be an issue now, but could become once called often (?)

rtyler: There is no significant cost overhead as far as I am aware with constructing "error" objects, which as far as I am aware are nothing special compared to other "objects" in JavaScript.

I know Java has overhead related to Exception construction, but to my understanding JavaScript has no similar performance penalties

rtyler:
We gotta find more JavaScripters :smile_cat:

PR #44, opened 2018-04-05
[JENKINS-50294] Prototype health checking

batmat:
Context: [JENKINS-50294](https://issues.jenkins-ci.org/browse/JENKINS-50294)

The goal here is to show (see the `tests.sh` addition) how health-checking can be done:
* simple check on /login that it returns HTTP-200
* simple health-check on the pre-configured metrics-plugin exposed URL


batmat:
Interesting, the last commit seems to have revealed a race condition because I don't think I really changed the logic.
I think the test just ran a wee bit faster previously and ran before the `healther-checker.log` log file was created (incorrectly) under `$JENKINS_HOME/logs`.

Need to fix/adapt metrics-plugin to https://issues.jenkins-ci.org/browse/JENKINS-50291

metrics-config.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version='1.1' encoding='UTF-8'?>
```
rtyler: Perhaps it would be a good time to add a new directory for Jenkins configuration files such as this, the log properties, yaml files, etc

metrics-config.xml
```diff
@@ -0,0 +1,15 @@
...
+<jenkins.metrics.api.MetricsAccessKey_-DescriptorImpl plugin="metrics@3.1.2.11">
+  <!-- TODO: check with Stephen if forcing "evergreen" as key is a bad bad idea
+  FIXME: seems like the "origins" field cannot be saved using the config UI, bug?
+  -->
+  <accessKeys>
+    <jenkins.metrics.api.MetricsAccessKey>
```
rtyler: Are these settings not configurable via the [Configuration as Code](https://github.com/jenkinsci/configuration-as-code-plugin) plugin? 

If not, I believe a ticket should be filed for the Metrics plugin, or for the Configuration as Code plugin, to add support.

batmat: Agreed.

metrics-config.xml
batmat: Yes. Will give it some thought.

metrics-config.xml
batmat: Filed https://github.com/jenkinsci/configuration-as-code-plugin/issues/169

rtyler: Yeah, the latter, please add a reference to the tickets as a comment in the XML when you move it to `configuration/`

batmat: Done @rtyler

batmat:
Will merge when the PR build succeeds. Or the first who sees this :P

PR #45, opened 2018-04-06
[JENKINS-50624] Rework the custom plugins builds system to make it easier to add/remove a plugin

batmat:
https://issues.jenkins-ci.org/browse/JENKINS-50624

So the idea is now, the `plugins-to-build.list` file contains a colon separated format that allows one to express the plugin we should build by ourselves.

```
# org:repo:branch
jenkinsci:metrics-plugin:master
jenkinsci:configuration-as-code-plugin:master
batmat:essentials-plugin:master
```

Also, then we do not need anymore to twiddle with the source file for `aria2c`. We just declare there everything need for a given plugin, then the custom/local build will be favored if present. Caveat: we might download a few plugins unnecessarily, but a plugin is generally so small it should not be an issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,14 +12,17 @@ all: check container
 lint: essentials.yaml shunit2
 	./tools/yamllint -s ./essentials.yaml
 	./tools/shellcheck -x tests/tests.sh
-	./tools/shellcheck -x scripts/shim-startup-wrapper.sh
+	./tools/shellcheck -x scripts/*.sh
 
 check: lint
 	$(MAKE) -C client $@
 	$(MAKE) -C services $@
 	$(MAKE) container-check
 
-container-prereqs: build/jenkins-support build/jenkins.sh scripts/shim-startup-wrapper.sh build/configuration-as-code/target/configuration-as-code.hpi build/essentials/target/essentials.hpi
+container-prereqs: build/jenkins-support build/jenkins.sh scripts/shim-startup-wrapper.sh build-plugins
+
+build-plugins:
+	./scripts/build-plugins.sh
 
 container-check: shunit2 ./tests/tests.sh container
 	./tests/tests.sh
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -12,14 +12,17 @@ all: check container
...
 
 check: lint
 	$(MAKE) -C client $@
 	$(MAKE) -C services $@
 	$(MAKE) container-check
 
-container-prereqs: build/jenkins-support build/jenkins.sh scripts/shim-startup-wrapper.sh build/configuration-as-code/target/configuration-as-code.hpi build/essentials/target/essentials.hpi
+container-prereqs: build/jenkins-support build/jenkins.sh scripts/shim-startup-wrapper.sh build-plugins
+
+build-plugins:
```
rtyler: This should be added as a PHONY target to the bottom of the file

scripts/build-plugins.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "hpi found under $local_path, no build. Wipe out if you wish to rebuild the plugin."
+  else
+    echo "Building $local_path (without tests...)"
+    cd "$local_path"
+    mvn --batch-mode -DskipTests clean package
+  fi
+done
+
+echo "Copying HPIs"
+find "$build_dir" -maxdepth 3 -name "*.hpi" -exec cp "{}" "$build_dir/.." \;
```
rtyler: Many plugins may generate other files marked `.hpi`, but I think that's only when tests are being run.

I did something similar for Code Valet  [here](https://github.com/CodeValet/master/blob/master/scripts/build-plugins#L77)

batmat: Yes. That is why I used `-maxdepth 3` to not find the .jpi/.hpi generated mainly under test (even if I build with `-DskipTests`)

Makefile
batmat: done

PR #50, opened 2018-04-10
Support storing information in the Status service

rtyler:
The Status service is largely responsible for the data storage and retrieval of information about a given Instance's "Status."

This means that it is only responsible for maintaining the "current state of the world" rather than driving any upgrade processes on the Instance per se.


rtyler:
Removing the work-in-progress label such that we can review and merge this code so long as it's passing.

Not sure it's worth holding half-finished branches in code review until everything is finalized

services/README.adoc
```diff
@@ -53,35 +55,48 @@ service will be receiving semi-arbitrary blobs of JSON from instances.
 [source]
 ----
     +------------+                           +----------------+
-    | Channel    |                           | Connection     |
+    | channels   |                           | connections    |
```
batmat: Are those tables? I feel it should be the contrary: make the others singular isn't it? It's what I've always seen and learned for RDBMS tables at least.

services/seeders/20180410172245-default-channels.js
```diff
@@ -0,0 +1,29 @@
...
+          name: 'beta',
+          createdAt: Sequelize.fn('NOW'),
+          updatedAt: Sequelize.fn('NOW')
+        },
+        {
+          name: 'general',
+          createdAt: Sequelize.fn('NOW'),
+          updatedAt: Sequelize.fn('NOW')
+        }
+    ], {});
```
batmat: Just food for thought, I guess you read it: https://www.chromium.org/getting-involved/dev-channel: 

> Stable channel: This channel has gotten the full testing and blessing of the Chrome test team, and is the best bet to avoid crashes and other issues.  It's updated roughly every two-three weeks for minor releases, and every 6 weeks for major releases.
Beta channel: If you are interested in seeing what's next, with minimal risk, Beta channel is the place to be.  It's updated every week roughly, with major updates coming every six weeks, more than a month before the Stable channel will get them.
Dev channel: Want to see what's happening quickly, then you want the Dev channel.  The Dev channel gets updated once or twice weekly, and it shows what we're working on right now.  There's no lag between major versions, whatever code we've got, you will get.  While this build does get tested, it is still subject to bugs, as we want people to see what's new as soon as possible.
Canary build: Canary builds are the bleeding edge. Released daily, this build has not been tested or used, it's released as soon as it's built.
Other builds: If you're absolutely crazy, you can download the latest working (and that's a very loose definition of working) build from download-chromium.appspot.com. You can also look for a more specific recent build by going to the Chromium continuous build waterfall, looking at the number near the top under "LKGR", and then going to this Google Storage bucket and downloading the corresponding build.

batmat: More specifically I think renaming `general` to `stable` over time might indeed convey a clearer message.

rtyler: @batmat I am aware of Chrome's stable channel, and I explicitly reject the label :smile: 

We are not exactly following Chrome's model because Chrome's model is not continuous, it's kind of a blend between a train and rolling release. The use of `stable` gives an implication that everything else is `unstable`, which is not the plan at all here. For something to get into the `stable` channel it must be blessed by a manual test team which we will never have for Jenkins Essentials (which isn't a bad thing).

services/README.adoc
rtyler: I'm not sure I understand this comment, is the problem with the diagram, or that table names are plurals?

I'm not sure I see the point in debating the merits of an extra character on table names :smile:

PR #52, opened 2018-04-11
Use /instance-identity/ instead for health-checking

batmat:
Align to https://github.com/batmat/jep/tree/6ea85ea540c341d3543fc819bf73632c0584e2a2/jep/0000

### DIFF
--- a/tests/tests.sh
+++ b/tests/tests.sh
@@ -101,9 +101,15 @@ test_essentials_telemetry_logging_is_found_on_disk() {
   assertEquals "0" "$?"
 }
 
-# JENKINS-50294 Health checking
+# not used for health-checking anymore, but kept for smoke testing
 test_login_http_200() {
   status_code=$( curl --silent --output /dev/null --write-out "%{http_code}" "http://localhost:$TEST_PORT/login" )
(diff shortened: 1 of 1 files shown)


tests/tests.sh
```diff
@@ -101,12 +101,18 @@ test_essentials_telemetry_logging_is_found_on_disk() {
...
-# JENKINS-50294 Health checking
+# not used for health-checking anymore, but kept for smoke testing
 test_login_http_200() {
   status_code=$( curl --silent --output /dev/null --write-out "%{http_code}" "http://localhost:$TEST_PORT/login" )
   assertEquals "0" "$?"
   assertEquals "200" "$status_code"
 }
+# JENKINS-50294 Health checking
+test_instance_identity_http_200() {
+  status_code=$( curl --silent --output /dev/null --write-out "%{http_code}" "http://localhost:$TEST_PORT/instance-identity/" )
```
jglick: Note that the JEP refers in a couple of cases to `/instance-identity`, which is wrong—that would be a redirect to `/instance-identity/`.

batmat: Fixed, thanks.

PR #55, opened 2018-04-12
WIP: test the /tmp/test file is written from the JSON logs sent to essentials.log.0

batmat:
Currently fails this way.

What I am trying to do: 
* set up the error logging callback
* write to the file to trigger the "transformation" callback (from one file to another currently)
* check the /tmp/test target file has the expected content

But I don't know if I'm on the right path or not. Will check again later today.

```shell
$ make check
# Checking to see if the directory exists because npm install updates the
# directory every time it runs, busting the GNU/Make cache causing rebuilds
if [ ! -d node_modules ]; then \
        ../tools/node npm install; \
fi;
../tools/node npm test

> evergreen-client@1.0.0 test /home/tiste/dev/github/jenkins-infra/evergreen/client
> npm run eslint && npm run jest


> evergreen-client@1.0.0 eslint /home/tiste/dev/github/jenkins-infra/evergreen/client
> eslint src/. test/. --config .eslintrc.json


> evergreen-client@1.0.0 jest /home/tiste/dev/github/jenkins-infra/evergreen/client
> jest

 PASS  test/auth.test.js

 RUNS  test/registration.test.js
 RUNS  test/error-telemetry.test.js
 RUNS  test/client.test.js
 RUNS  test/auth.test.js

 RUNS  test/registration.test.js
 RUNS  test/error-telemetry.test.js
 RUNS  test/client.test.js

 FAIL  test/error-telemetry.test.js
  ● Console

    console.log lib/__mocks__/fs.js:3
      Using a memfs filesystem...

  ● Error Telemetry Logging › setup() call › writing to essentials logging file should forward to another

    assert.equal(received, expected) or assert(received)

    Expected value to be (operator: ==):
      true
    Received:
      false

      38 |       setTimeout(function() {
      39 |         logger.error('YAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA');
    > 40 |         assert(fs.existsSync('/tmp/test'));
      41 |         assert.equal('MESSAGE =WAT\nMESSAGE=WAT2', fs.readFileSync('/tmp/test','utf8'));
      42 |       }, 5000);
      43 |       jest.runAllTimers();

      at test/error-telemetry.test.js:40:9
      at Object.it (test/error-telemetry.test.js:43:12)

 PASS  test/client.test.js

 RUNS  test/registration.test.js

Test Suites: 1 failed, 2 passed, 3 of 4 total
Tests:       1 failed, 3 passed, 4 total
Snapshots:   0 total
Time:        1sinfo: Saving uuid to disk at /evergreen/keys/uuid.json
 PASS  test/registration.test.jsout a private or public key on Registration
  ● Console

    console.log lib/__mocks__/fs.js:3
      Using a memfs filesystem...


Test Suites: 1 failed, 3 passed, 4 total
Tests:       1 failed, 21 passed, 22 total
Snapshots:   0 total
Time:        1.885s
Ran all test suites.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! evergreen-client@1.0.0 jest: `jest`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the evergreen-client@1.0.0 jest script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/node/.npm/_logs/2018-04-12T15_51_12_866Z-debug.log
npm ERR! Test failed.  See above for more details.
make: *** [Makefile:6: check] Error 1
```


batmat:
@rtyler if you have a few minutes to see and possibly propose some way or point out the mistakes I did, that would be great. Thanks!

client/test/error-telemetry.test.js
```diff
@@ -0,0 +1,48 @@
...
+      mkdirp.sync(logsDir);
+      fs.writeFileSync(logsDir + 'essentials.log.0', '{"timestamp":1523451065975,"level":"SEVERE","message":"WAT"}');
+
+      const response = (new ErrorTelemetry()).setup();
+      assert(response instanceof Promise);
+
+      // Write again
+      fs.appendFileSync(logsDir + 'essentials.log.0', '{"timestamp":1523451065975,"level":"SEVERE","message":"WAT2"}');
+
+      assert(!fs.existsSync('/tmp/test'));
```
Wadeck: Could be a problem if the Tail event is triggered between the appendFileSync and this call. Perhaps better to place it before the line 34.

client/test/error-telemetry.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+      setTimeout(function() {
+        logger.error('YAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA');
+        assert(fs.existsSync('/tmp/test'));
+        assert.equal('MESSAGE =WAT\nMESSAGE=WAT2', fs.readFileSync('/tmp/test','utf8'));
+      }, 5000);
+      jest.runAllTimers();
```
Wadeck: Because of this call, the timeout is executed directly and so you fall in a concurrency problem I think. Your `ErrorTelemetry` method does not have the time to do its stuff, especially the Tail that seems to use an interval meaning it's potentially not real time.

Option 1) you can edit the code and place something like the following to know if it's really a concurrency problem or not:
```
// in line 22, replace the arrow function by
(done) => {

// instead of lines 38-43
while(!fs.existsSync('/tmp/test');
assert.equal('MESSAGE=WAT\nMESSAGE=WAT2', fs.readFileSync('/tmp/test','utf8'));
done();
```
With that, the test will wait the file exists and only at that point it will check the content. If the file will never exist, the test will timeout itself. The `done` tells the script runner that the test is asynchronous and it must wait until the method is called.

Option 2) do not use the jest mocked timers to have a real 5sec wait.

client/test/error-telemetry.test.js
```diff
@@ -0,0 +1,48 @@
...
+const logger = require('winston');
+
+describe('Error Telemetry Logging', () => {
+  beforeEach(() => {
+    /* Make sure memfs is flushed every time */
+    fs.volume.reset();
+  });
+
+  describe('setup() call', () => {
+    it('should not return a Promise', () => {
```
Wadeck: in your case it's "should return a Promise"

but I am not sure that returning a Promise is a good idea in this case (in the current version of the code) because you are not waiting for other Promise, your code add listeners. If you simply return void, the listeners will still "live" and do the stuff you want.

In general, you "async" a method if you want to use some "await" calls inside its body.

client/test/error-telemetry.test.js
```diff
@@ -0,0 +1,48 @@
...
+      const response = (new ErrorTelemetry()).setup();
```
Wadeck: nit: parenthesis not required

client/test/error-telemetry.test.js
batmat: @Wadeck thanks for the help.

I cannot make any of those work. If I try what you propose with the loop above, it seems to be blocked and never return.

I suspect this is due to the Javascript threading model and using an active loop is just keeping it looping, but nothing else can move forward while we're looping (?)

PR #56, opened 2018-04-12
Starting to incorporate the use of the Status service into the client

rtyler:
There are still a lot of commits to come into this pull request, but slowly moving towards sending Status information on some periodic trigger within `evergreen-client`.


rtyler:
@batmat is out today. I'm going to land this and keep fixing stuff, easy to revert or update code later :smile:

PR #60, opened 2018-04-17
Add first error telemetry logging prototyping

batmat:
The client sets up handler to watch the `essentials.log.0` file and
send them to the disk currently under /tmp/test.

It is *obviously* not the end goal, but aims at showing we can watch that
file and send things from there to somewhere else.

Thanks a lot to @wadeck for his great help making this work.

Supersedes https://github.com/jenkins-infra/evergreen/pull/55


client/src/lib/error-telemetry.js
```diff
@@ -0,0 +1,68 @@
...
+
+class ErrorTelemetry {
+  constructor(app, options) {
+    this.app = app;
+    this.options = options;
+  }
+
+  setup() {
+    logger.info('Setting up error logging...');
+    var loggingFile = this.fileToWatch();
```
rtyler: Generally I've been avoiding the use of `var` in favor of the block-scoped [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let).

FWIW

client/src/lib/error-telemetry.js
```diff
@@ -0,0 +1,68 @@
...
+
+
+    // FIXME: this is wrong, we should probably /just/ retry, but handling this an error
+    // BTW, we should define a way to also report critical errors like those to the backend
```
rtyler: I would expect this mechanism for reporting critical errors to be within the scope of [the error telemetry service API contract](https://issues.jenkins-ci.org/browse/JENKINS-49811)

client/src/lib/error-telemetry.js
```diff
@@ -0,0 +1,68 @@
...
+    // The logging file being absent, or some critical parts being wrong on the Jenkins instance,
+    // basically nothing should never make the client crash I think
+    // (if a connection is established, sure)
+    if(!fs.existsSync(loggingFile)) {
+      throw new Error('logging file ' + loggingFile + ' not found. ');
+    }
+
+    const tail = new Tail(loggingFile, {
```
rtyler: Would there be any utility in making `tail` an object variable, e.g. `this.tail`? Will future methods need to access this tail object?

client/src/lib/error-telemetry.js
```diff
@@ -0,0 +1,68 @@
...
+      fromBeginning: true
+    });
+
+    tail.on('line', data => {
+      const json = JSON.parse(data);
+
+      const text = 'MESSAGE='+json.message; //, date="+new Date(json.timestamp) +'\n';
+      fs.appendFile(TARGET, text+'\n', err => {
```
rtyler: After you were the one who reminded me about template strings in ES6, you should use them here :smiling_imp:

client/src/lib/error-telemetry.js
```diff
@@ -0,0 +1,68 @@
...
+      });
+    });
+
+    tail.on('error', error => {
+      logger.error('ERROR: ' + error);
+    });
+  }
+
+  fileToWatch() {
+    var path = '';
```
rtyler: This can be simplified a bit, which I would recommend, with:

```javascript
let path = process.env.ESSENTIALS_LOG_FILE;

if (!path) {
  path = 'some/default/log';
}
```

client/test/error-telemetry.test.js
```diff
@@ -0,0 +1,45 @@
...
+describe('Error Telemetry Logging', () => {
+  beforeEach(() => {
+    /* Make sure memfs is flushed every time */
+    fs.volume.reset();
+  });
+
+  describe('setup() call', () => {
+
+    // FIXME: only hackish, the end goal is definitely not to forward to another file
+    it('writing to essentials logging file should forward to another', done => {
```
rtyler: From a test structure perspective this is a bit hackish too. Please try to organize setup-type tasks into a `beforeEach` block rather than inside the `it` blocks

Generally I have been trying to keep the `it` blocks to only a few lines which include the _assertions_ rather than the entirety of the set up.

`beforeEach` for example, can be nested inside of nested `describe` blocks FWIW

client/test/error-telemetry.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+      assert(fs.existsSync(logFile));
+
+      mkdirp.sync('/tmp');
+      const response = new ErrorTelemetry().setup();
+      assert(!(response instanceof Promise));
+
+      assert(!fs.existsSync('/tmp/test'));
+
+      setTimeout( () => {
```
rtyler: I'm not sure how jest works with `setTimeout`, you may want to consider using [Timer mocks](https://facebook.github.io/jest/docs/en/timer-mocks.html)

Although, I'm not sure if that's necessary for this initial kind of hacky test case

client/src/lib/error-telemetry.js
Wadeck: In this case it's even a `const` as it's never modified after that.

client/src/lib/error-telemetry.js
Wadeck: or even 
```
const path = process.env.ESSENTIALS_LOG_FILE || '/evergreen/jenkins/var/logs/essentials.log.0';
```

but in such case you loose the log part that could be interesting.

client/package-lock.json
```diff
@@ -6358,6 +6358,11 @@
         "string-width": "2.1.1"
       }
     },
+    "tail": {
+      "version": "1.2.3",
+      "resolved": "https://registry.npmjs.org/tail/-/tail-1.2.3.tgz",
+      "integrity": "sha1-sI1vp5+5KIaWMaNBpRwUSXwcQlU="
+    },
```
Wadeck: are you using a version of NPM (5+) that takes care of that file in production ?

it's just to be sure, to avoid security issues due to dangerous versions

client/src/lib/error-telemetry.js
batmat: I've fixed this and enabled the `no-var` rule in ESLint, see latest commits.

client/src/lib/error-telemetry.js
batmat: Possibly, but YAGNI for now IMO. Easy enough to change if so.

client/src/lib/error-telemetry.js
batmat: Oh I actually incorrectly thought the eslint rule 

```json
"quotes": [
      "error",
      "single"
    ],
```

was forbidding this, fixing.

client/test/error-telemetry.test.js
batmat: Yeah, and I've tried quite a bit. Possibly I missed something, but couldn't make it work for now. We ended up removing them when hacking with @Wadeck mentoring.

I think I'll try to reintroduce them when I'm more used to the wholee thing.

client/test/error-telemetry.test.js
batmat: ack will do

batmat:
Ready for review. 

Next step is to implement service and call it instead of just write to another file.

batmat:
@rtyler yes, will do. Was planning to adjust progressively to match the JEP, as described in https://github.com/jenkinsci/jep/tree/master/jep/304#summary : 

>  ... it is suggested to: 
> * store the last timestamp that was read ;

PR #64, opened 2018-04-24
Remove the channels model and introduce updates and update levels

rtyler:
Fairly straightforward work associated with this JEP:

https://github.com/jenkinsci/jep/pull/89


services/README.adoc
```diff
@@ -54,38 +54,49 @@ service will be receiving semi-arbitrary blobs of JSON from instances.
...
+                                            |   +-----------------+
+                                            |   |  id             |
+                                            +---+ *uuid           |
+                                                |  lastConnected  |
+                                                |  createdAt      |
+                                                |  updatedAt      |
+                                                +-----------------+
 
 * denotes an index
+`id` fields are always considered primary keys
```
batmat: I don't think it renders as you expected it using \`id` as it's in the global preformatted source block.

services/README.adoc
```diff
@@ -54,38 +54,49 @@ service will be receiving semi-arbitrary blobs of JSON from instances.
...
+                           |  timezone  |   ^^
+                           |  flavor    |   ||
+  +----------------+   +---+  updateId  |   ||
+  | updates        |   |   |  createdAt |   ||
+  +----------------+   |   |  updatedAt |   ||
+  | id             <---+   +------------+   ||
+  | commit         |                        ||
+  | channel        |                        ||  +---------------------+
+  | manifest (JSON)|                        ||  | capabilities        |
+  | tainted        |                        ||  +---------------------+
```
batmat: as said on the JEP, I wonder if this should be a richer field than just a boolean.

PR #67, opened 2018-04-27
Pull the JWT configuration from the config files rather than hard coded

rtyler:


client/src/client.js
```diff
@@ -46,7 +46,12 @@ class Client {
     this.app.configure(restClient.fetch(fetch));
 
     // FIXME: move this into runloop once we figure out how to run it locally during testing
-    this.errorTelemetry.setup();
+    try {
+      this.errorTelemetry.setup();
```
batmat: Why not, but so you had issues with this I'm curious? I didn't AFAICT, and CI was not barking either.

services/src/app.js
```diff
@@ -66,18 +66,19 @@ else {
 app.hooks(appHooks);
 
 /* Configure the authentication provider via @feathersjs/authentication-jwt and
- * passport-jwt (https://github.com/themikenicholson/passport-jwt)
+ *` passport-jwt (https://github.com/themikenicholson/passport-jwt)
```
batmat: ?

PR #69, opened 2018-04-30
Properly run both client and server containers from a `make run`

rtyler:
When executing `make run` in the root directory, everything will be checked,
built, and then run. After which bigger integration testing could operate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,8 @@ services:
     image: postgres:alpine
     environment:
       - 'POSTGRES_PASSWORD=grassisevergreener'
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -14,3 +16,8 @@ services:
     depends_on:
       - db
 
+  client:
+    image: jenkins/evergreen
+    build: ./
+    depends_on:
```
batmat: Note to myself: need to add:

```yaml
environment:
      - 'EVERGREEN_ENDPOINT=http://backend:3030/'
```

PR #74, opened 2018-05-14
Ensure that a database dump occurs before running tests

rtyler:
We need this to be able to seed the database properly for integration tests


rtyler:
The Pipeline is continuing to fail for largely expected reasons. Merging this to improve `update-levels` for now.

PR #77, opened 2018-05-15
[JENKINS-51345] Save the essentials.log.0 file after test run

batmat:
image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -50,7 +50,7 @@ pipeline {
             }
             post {
                 always {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -50,7 +50,7 @@ pipeline {
             }
             post {
                 always {
-                    archiveArtifacts artifacts: 'build/tests-logs*/**.log'
+                    archiveArtifacts artifacts: 'build/tests-logs*/**.log*'
```
rtyler: This is no longer going to work btw because the directory was changed in the `tests/utilities` file

PR #78, opened 2018-05-15
Store update level 50951

rtyler:
This pull request also merges `master` and `update-levels` changes, which should be rebased in the near future.

The tests are going to fail because `essentials.log.0` has nothing in it, which I believe is the correct behavior.


rtyler:
Going to merge this, the `update-levels` branch is going to need to be rebased before merge

PR #80, opened 2018-05-15
Implement [hard-coded] Update Levels and client/server interactions

rtyler:
This pull request brings a lot of work to master to make the Client and Server both behavior in a manner more resembling the Update Levels described in [JEP-307](https://github.com/jenkinsci/jep/blob/master/jep/307)


client/test/periodic.test.js
```diff
@@ -13,6 +13,13 @@ describe('The periodic module', () => {
     });
   });
 
+  describe('runDaily()', () => {
+    it('allows registration of a daily callback', () => {
+      let p = periodic(app);
+      assert.ok(p.runDaily('jest-fun', () => { }));
```
batmat: IIUC this does not really test the daily run, just that calling the method does not fail.

PR #86, opened 2018-05-23
 [JENKINS-51469] Move Metrics config under new "unclassified" root

batmat:
[JENKINS-51469](https://issues.jenkins-ci.org/browse/JENKINS-51469)

See also https://github.com/jenkinsci/configuration-as-code-plugin/pull/216

### DIFF
--- a/config/jenkins-configuration.yaml
+++ b/config/jenkins-configuration.yaml
@@ -2,6 +2,8 @@
 jenkins:
   systemMessage: "Welcome to Jenkins Essentials!"
   numExecutors: 0
+
+unclassified:
   metricsaccesskey:
     accessKeys:
       - key: "evergreen"
--- a/services/src/services/update/update.hooks.js
+++ b/services/src/services/update/update.hooks.js
@@ -84,7 +84,7 @@ class UpdateHooks {
             }
           },
           {
-            'url' : 'http://mirrors.jenkins-ci.org/plugins/configuration-as-code/0.5-alpha/configuration-as-code.hpi',
+            'url' : 'https://repo.jenkins-ci.org/incrementals/io/jenkins/configuration-as-code/0.7-alpha-rc240.6e7dd119b0d3/configuration-as-code-0.7-alpha-rc240.6e7dd119b0d3.hpi',
             'checksum' : {
               'type' : 'sha256',
               'signature' : '703f04ad2820a39e131ec66de6872c13556f6a7345e3f50471c2ff7bf3d19f2c'


config/jenkins-configuration.yaml
```diff
@@ -2,6 +2,8 @@
 jenkins:
   systemMessage: "Welcome to Jenkins Essentials!"
   numExecutors: 0
+
+unclassified:
```
rtyler: I haven't been able to follow all the pull request discussions about this on jenkinsci/configuration-as-code-plugin.

I haven't seen this documented anywhere, so perhaps a comment in this file to explain when this might become "unclassfieid?" Everything else in the YAML file is fairly self explanatory or understandable to me, but the special distinction of this tree of configuration is a bit wacky to me :slightly_frowning_face:

batmat: @rtyler IIUC, that new root got introduced as a default for plugins that do not specify an explicit categorization they should be put in.

Coming from the core `@Symbol` value:
https://github.com/jenkinsci/jenkins/blob/79ec930a8ef4f12d2f2574c50c2f42272e5c75fb/core/src/main/java/jenkins/model/GlobalConfigurationCategory.java#L62

Can you possibly give a bit more precisions @ndeloof @ewelinawilkosz? Thanks!

ndeloof: as I implemented yaml generation from a live instance, I wasn't able to distinguish legitimate Jenkins instance attribute vs all the orphaned descriptors we used to attach to it. On the other hand, we already had moved all well categorized descriptors to dedicated root elements (`tools`, `security`...). So I remove the special handling of those uncategorized descriptors to root.

rtyler (approved):
I'm disappointed that this is likely to break as soon as a `@Symbol` annotation is added to the Metrics plugin.

Doesn't seem like there is a good solution around that unless we decide to stop using Configuration as Code

ndeloof:
adding Symbol to metrics will change the tag name, having Metrics to declare a category will change where the node get attached in yaml tree. Configuration-as-Code is based on the model exposed by Jenkins components, just like the REST API does. Upgrading a component in this model never is transparent.

batmat:
@ndeloof IMO, it is `configuration-as-code` mission to design how this kind of change can be made backward-compatible. I'm obviously *not* saying CasC should be able to accomodate all rogue changes and releases from plugins out there, but that there should be a documented way to, for instance in that case, move from a category to another, and that things keep working (at least for some time).

Failing that, to my disappointment, it will be hard to convince more and more people to invest time, and migrate their current way of configuring their instances to CasC.

For example, maybe we _could_ consider something like:

If one goes from 
```java
@Extension
public static class MyConfiguration extends GlobalConfiguration  { ... }
```

```java
@Symbol("oh-my")
@Extension
public static class MyConfiguration extends GlobalConfiguration  { ... }
```
Which would as-of currently make CasC fail?

Then, on CasC front, CasC could possibly:
* if not found, try to find `MyConfiguration` under a more specific one than `unclassified`. 
* And once found, spit a big fat warning?

If needed, I suppose we could introduce things like `@PreviousSymbol` to provide a way for plugin developers to change things while keeping users [happy]?

ndeloof:
This would indeed let CasC report a configuration mismatch.

configuration-as-code design is to generate a schema from a live instance. It can at same time generate adequate tooling to validate a configuration vs a target core/plugins version set to detect mismatch. We _could_  also generate a partial schema per plugin so a plugin release note can include reference on changes for CasC users.

About your suggestion, this sounds something feasible for this simple use-cases, as we still can compute the previous component's inferred symbol, as a lower priority candidate. But it will be limited to such minor changes. Also this would send a bad signal that configuration is still valid while it require an update. Sounds like we will get back to this growing set of `@Deprecated` stuff in jenkins codebase.

PR #87, opened 2018-05-23
[JENKINS-51496] secure error telemetry service

batmat:
[JENKINS-51496](https://issues.jenkins-ci.org/browse/JENKINS-51496)

### DIFF
--- a/client/src/lib/error-telemetry.js
+++ b/client/src/lib/error-telemetry.js
@@ -25,19 +25,19 @@ class ErrorTelemetry {
   * (Private) default behaviour for the output where to send data to when the watched logging file
   * has a modification detected.
   */
-  callErrorTelemetryService(app, logDataObject, uuid, token) {
+  callErrorTelemetryService(app, logDataObject) {
 
     const api = app.service('telemetry/error');
 
     const payload = {
       log: logDataObject,
-      uuid: uuid
+      uuid: this.uuid
     };
     return api.create(
       payload
       ,
       {
-        headers: { Authorization: token }
+        headers: { Authorization: this.token }
       }
     ).then((res) => {
       logger.info('pushed as ', res);
@@ -50,7 +50,7 @@ class ErrorTelemetry {
   * monitoredFile: path to the log file to watch
   * outputFunction(app,line): the function that will be called on each new line detected
   */
-  setup(monitoredFile, outputFunction=this.callErrorTelemetryService) {
+  setup(monitoredFile) {
     logger.info('Setting up error logging...');
 
     let loggingFile = '';
@@ -75,7 +75,7 @@ class ErrorTelemetry {
       logger.debug('Reading line:', data);
 
       try {
-        outputFunction(this.app, JSON.parse(data), this.uuid, this.token);
+        this.callErrorTelemetryService(this.app, JSON.parse(data));
       } catch(err) {
         logger.error(`Unable to parse as JSON, corrupt log line? ***${data}***`, err);
       }
--- a/client/src/lib/error-telemetry.js
+++ b/client/src/lib/error-telemetry.js
@@ -15,27 +15,42 @@ class ErrorTelemetry {
     this.options = options;
   }
 
+  authenticate(uuid, token) {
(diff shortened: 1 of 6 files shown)


batmat:
The only addition for now is https://github.com/jenkins-infra/evergreen/pull/87/commits/100d2fed2654d56c6cb47596007bc5e718b0067a which adds a test to show error telemetry was not yet secured until this fix.

batmat:
Failed as expected:
```
test_error_telemetry_service_is_secured
ASSERT:Service should return 401 expected:<401> but was:<201>
shunit2:ERROR test_error_telemetry_service_is_secured() returned non-zero return code.
```

Just pushed the fix as https://github.com/jenkins-infra/evergreen/pull/87/commits/8337322c5acdd5efcfa0696f6aeafa8f13e4bafc

batmat:
OK, this now fails because of the coverage threshold. Will fix tomorrow.

batmat:
Still ready for review I think @rtyler if you have a bit of time to look at it.

client/src/lib/error-telemetry.js
```diff
@@ -15,27 +15,42 @@ class ErrorTelemetry {
     this.options = options;
   }
 
+  authenticate(uuid, token) {
+    logger.info(`Error telemetry: uuid=${uuid}, token=${token}`);
```
rtyler: Please remove this log statement. I don't believe it's adding any value at this point

client/src/lib/error-telemetry.js
```diff
@@ -15,27 +15,42 @@ class ErrorTelemetry {
...
+
   /**
-   * (Private) default behaviour for the output where to send data to when the watched logging file
-   * has a modification detected.
-   */
-  callErrorTelemetryService(app, logDataObject) {
+  * (Private) default behaviour for the output where to send data to when the watched logging file
+  * has a modification detected.
+  */
+  callErrorTelemetryService(app, logDataObject, uuid, token) {
```
rtyler: I'm not sure I understand why this is being called with attributes that are already members of the objects? Why isn't the function referring to `this.uuid` and `this.token`?

client/src/lib/error-telemetry.js
```diff
@@ -15,27 +15,42 @@ class ErrorTelemetry {
...
   }
 
   /**
-   * monitoredFile: path to the log file to watch
-   * outputFunction(app,line): the function that will be called on each new line detected
-   */
+  * monitoredFile: path to the log file to watch
+  * outputFunction(app,line): the function that will be called on each new line detected
+  */
   setup(monitoredFile, outputFunction=this.callErrorTelemetryService) {
```
rtyler: What I imagine the answer to my previous question about updating the prototype of `callErrorTelemetryService`, I think this class needs to be refactored a bit to disallow this unnecessary passing of an arbitrary `outputFunction` here. 

I thought this was originally meant as a kind of dependency injection for testing, but that doesn't appear the case.

I think this object should be "sealed up" a bit, and the `outputFunction` is an implementation detail that no callers should bother with.

client/src/lib/error-telemetry.js
```diff
@@ -61,9 +76,9 @@ class ErrorTelemetry {
       logger.debug('Reading line:', data);
```
rtyler: If some of the reasoning behind `outputFunction` is to scope into this callback, then I suggest looking at the re-assignment of `this` to `self` which is used in other places for just this purpose

services/src/services/errorTelemetry/errorTelemetry.hooks.js
```diff
@@ -31,7 +35,8 @@ class ErrorTelemetryHooks {
                 throw new errors.BadRequest(`Missing required field '${field}'`);
               }
             });
-          }
+          },
+          ensureMatchingUUID
```
rtyler: Please make this the first hook. Whenever we have authenticated APIs like this which take a `uuid` parameter, I would like the ensure hook to be first in the list to avoid any other hooks needing to do potentially unnecessary work (and resulting in work being done on an invalid request, there are some security concerns that raises for me which don't necessarily apply in this context)

PR #89, opened 2018-05-24
fix make check on OSX

ndeloof:


tools/node
```diff
@@ -2,14 +2,16 @@
...
+tty -s
 if [ $? -eq 0 ]; then
     TTY_ARGS="-ti"
 fi;
 
 exec docker run --net host --rm ${TTY_ARGS} \
     -u $(id -u):$(id -g) \
     -w ${PWD} \
+    --mount type=tmpfs,destination=/.npm \
+    --mount type=tmpfs,destination=/.config \
```
ndeloof: maybe those should be named volumes ? So they can be reused between runs

batmat: Just tested locally, it does not feel like the perf difference would be big, right? 
I'm not feeling strongly about that. For now it seems to work pretty well. 

I do not know Node enough to measure what is the risk a stale `.config` or `.npm` gets in the way at some point.

I guess we can keep it this way until it proves necessary to do differently.

PR #92, opened 2018-05-28
[JENKINS-51448] First P1 use cases identification

batmat:
[JENKINS-51448](https://issues.jenkins-ci.org/browse/JENKINS-51448)

Can be reviewed already. This PR can also serve to discuss those cases, and find new ones if deemed missing. 

*What is missing is the risk analysis to be done for each case. There are already a few lines written, and it's still very much WIP.*

### DIFF
--- a/docs/USE-CASES.adoc
+++ b/docs/USE-CASES.adoc
@@ -1,0 +1,35 @@
+= _Essentials_ Use Cases
+:toc:
+:sectnums:
+
+This document summarizes the Priority 1 use cases for _Jenkins Essentials_.
+
+NOTE: This is still early work.
+To help the thought process, the cases definition and analysis was done by thinking about what a demonstration would need to go through.
+
+== Priority 1 Cases: Risk analysis
+
+=== I can start an instance from scratch and it's available
+
+* Registration & authentication service need to be available:
+** An Internet connection is required
+** Bandwidth needs to be enough to download things in a _reasonable_ time.
+(It can probably be more than 5 minutes, if there is an issue.)
+* Binaries repository must be available.
+
+=== Jenkins is constantly upgraded safely
+
+* If the backend becomes unavailable:
+** Check the evergreen-client does reconnect automatically when back.
+* it the unavailability happens just after a failed upgrade, can we rollback offline?
+
+=== Starting in a given cloud environment, Essentials is auto-configured and ready to build with it
+
+Concrete examples:
+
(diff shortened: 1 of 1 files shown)


docs/USE-CASES.adoc
```diff
@@ -0,0 +1,35 @@
...
+== Priority 1 Cases: Risk analysis
+
+=== I can start an instance from scratch and it's available
+
+* Registration & authentication service need to be available:
+** An Internet connection is required
+** Bandwidth needs to be enough to download things in a _reasonable_ time.
+(It can probably be more than 5 minutes, if there is an issue.)
+* Binaries repository must be available.
+
```
raul-arabaolaza: Missing something about how to deal with network errors: 
* Are the operations needed for first start and setup atomic? 
** If not, in case of network error, does the operations resume? Are there any sanity checks of the downloaded resources?
* Does the first start ad configuration have a default timeout?

docs/USE-CASES.adoc
```diff
@@ -0,0 +1,35 @@
...
+=== Jenkins is constantly upgraded safely
+
+* If the backend becomes unavailable:
+** Check the evergreen-client does reconnect automatically when back.
+* it the unavailability happens just after a failed upgrade, can we rollback offline?
+
```
raul-arabaolaza: ditto, is not only about reconnecting, but also making sure you are in a safe state.

+1000 for the offline rollback

docs/USE-CASES.adoc
batmat: Wow, time flies. 

So, yeah @raul-arabaolaza as I tried to explain in the PR summary above, but I realize I was unclear, I was actually mainly here focusing on identifying the Use Case. I was planning to dig more into risk analysis in a followup PR.

docs/USE-CASES.adoc
```diff
@@ -0,0 +1,35 @@
+= _Essentials_ Use Cases
+:toc:
+:sectnums:
+
+This document summarizes the Priority 1 use cases for _Jenkins Essentials_.
+
+NOTE: This is still early work.
+To help the thought process, the cases definition and analysis was done by thinking about what a demonstration would need to go through.
```
batmat: This has been going for quite long. So as there is this notice, going to merge it as at least the core intent of this (the use cases themselves, not their risks all analyzed) seems to be fulfilled.

PR #95, opened 2018-05-30
[JENKINS-49866] Automatically configure Docker Cloud if a Docker host is available

batmat:


config/supervisord.conf
```diff
@@ -23,3 +24,9 @@ stdout_logfile_maxbytes=0
 redirect_stderr=true
 startsecs=10
 startretries=0
+user=jenkins
+
+
+[program:socat]
+command=socat -d -d TCP4-LISTEN:2375,fork UNIX-CONNECT:/var/run/docker.sock
```
rtyler: I'm so confused :smile_cat:

batmat:
Heh. No guarantee you are going to like the solution I found. But that is
one at least. More coming soon :).

Le mer. 30 mai 2018 à 17:48, R. Tyler Croy <notifications@github.com> a
écrit :

config/as-code/docker-cloud.yaml
```diff
@@ -0,0 +1,17 @@
...
+            uri: "tcp://localhost:2375/"
+        templates:
+          - labelString: "agent"
+            dockerTemplateBase:
+              image: "jenkins/slave"
+            remoteFs: "/home/jenkins/agent"
+            connector:
+              attach:
+                user: "jenkins"
+            instanceCapStr: "10"
```
pjdarton: Unless you're sure you need an instance cap, I'd suggest that you leave it blank.
If you leave it blank then Jenkins will (or should) start as many (of those) containers as it needs to.

If you're concerned about overloading the local docker engine then I'd suggest that setting a cloud-level cap would be the right thing to do.

batmat: @pjdarton though I didn't put that with a strong thinking behind, I would be more leaning toward putting _something_, to somehow offer an acceptable default behaviour for most setups.

(Like: running 1000 containers looks definitely too high and would kill the master itself on most setups, 100? 20?
Also, I do plan to limit the memory per container, though same: not sure what value to put).

Anyway, thanks a bunch for reviewing this. It's really appreciated, and I'll make sure to put you in the loop for upcoming changes about this (and anything else if you're interested :-)).

pjdarton: FYI we use docker a lot and we don't bother specifying a per-container memory limit.
Different usage scenarios will need *very* different values - some containers can run in a megabyte and some require over a gigabyte, and users of may be shocked that anyone could choose such a small/large number that's so very different to their needs.  Docker itself defaults to not specifying a limit for this very reason.
We just limit the number of containers we run on each docker host, although for our smaller docker hosts we limit it to 6 (as our containers each run IBM db2, and that is not a lightweight bit of software).

If you want to specify a default limit then how about setting a per-cloud limit of 4 containers?
That's small enough that it shouldn't stress most docker daemons, but large enough to demonstrate that we can have multiple containers running at once.
By comparison, Jenkins defaults to having only two executors on the master, so a limit of 4 containers is showing twice that capacity.

batmat: I think at that stage we should slightly conservative, and so yes I like the idea to set the limit at the cloud level to 4 or something.
 
Same goes for container memory: I do want us to be conservative for now, and open up as we better understand things and get feedback from users. "Closing" more features afterwards is as always more difficult IMO.
My idea would be to define "something", but the hard part comes indeed when defining actual values :)

As to:

> If you want to specify a default limit then how about setting a per-cloud limit of 4 containers?

IIUC this would mean removing the `instanceCapStr: "10"` line under `templates`, and use `containerCap: 4` directly under the `docker` node right @pjdarton ?

rtyler:
@batmat I think a conflict exists because of the hard-coded changes for the Update Levels in master, once you rebase feel free to merge

PR #98, opened 2018-06-01
Check ingest.yaml with yamllint too, and fix it

batmat:
Well, "fix" it. As the line-length made no sense to me to scavenge the config file for, I preferred to bump the 1950's default of 80 to 100.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,6 @@
+---
+rules:
+  line-length:
+    max: 100
+    allow-non-breakable-words: true
+    allow-non-breakable-inline-mappings: false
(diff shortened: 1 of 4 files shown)


.yamllint
```diff
@@ -0,0 +1,6 @@
+---
+rules:
+  line-length:
+    max: 100
+    allow-non-breakable-words: true
+    allow-non-breakable-inline-mappings: false
```
batmat: only the `max` line I wanted to change from the default. But the tool requires all fields to be specified when a given rule (here `line-length`) is specified. :man_facepalming:

PR #100, opened 2018-06-05
POSTing ingest.yaml for creating update levels

rtyler:
See JENKINS-50737


services/src/services/status/status.hooks.js
```diff
@@ -24,18 +25,21 @@ class StatusHooks {
         ],
         create: [
           ensureMatchingUUID,
+          dbtimestamp('createdAt'),
```
batmat: funny, I thought that was OOTB, OK.

PR #102, opened 2018-06-07
incorporating feathers-swagger for documenting APIs

rtyler:
Still a work in progress while I get the rest of the models documented and defined.

I think this is going to be very helpful and important for documenting our APIs more clearly than the current "just read the source" approach :smiling_imp: 

Example Swagger UI:

selection_430


rtyler:
The test failures are there now because the coverage threshold was lowered; will address in the morning

batmat:
Looks nice, will test it locally after lunch.

PR #105, opened 2018-06-11
Serve computed update manifests to clients

rtyler:
This pull request got a bit bigger than I wanted it to, but :man_shrugging: 

This implements support for processing `ingest.yaml` and serving that up appropriately to clients. What we're missing in order to make a proper demo is a version of `ingest.yaml` which represents the same thing that the 'terrible hard coded thing" used to do


services/src/services/update/update.class.js
```diff
@@ -0,0 +1,177 @@
...
+    super(options);
+
+    this.app = options.app;
+
+    /*
+     * Undefining some APIs which we don't want/need.
+     */
+    this.update = undefined;
+    this.patch = undefined;
+    this.remove = undefined;
```
batmat: This will work, but AFAICT this does not seem like the idiomatic way to disable some methods on a given service.
Seems like this is done through https://feathers-plus.github.io/v1/feathers-hooks-common/#disallow

i.e. using hooks, copying for possible future ref in case the link above dies:

```javascript
const { disallow, iff } = require('feathers-hooks-common');
module.exports = { before: {
    // Users can not be created by external access
    create: disallow('external'),
    // A user can not be deleted through the REST provider
    remove: disallow('rest'),
    // disallow calling `update` completely (e.g. to allow only `patch`)
    update: disallow(),
    // disallow the remove hook if the user is not an admin
    remove: iff(context => !context.params.user.isAdmin, disallow())
} };
```

services/src/services/update/update.class.js
```diff
@@ -0,0 +1,177 @@
...
+      };
+
+      this.prepareManifestFromRecord(record, computed);
+      /*
+       * Last but not least, make sure that any flavor specific updates are
+       * assigned to the computed update manifest
+       */
+      this.prepareManifestWithFlavor(instance, record, computed);
+
+      if (latestVersion.length === 1) {
```
batmat: Because of the `$limit` above, I guess this `if` shouldn't even be here because it will always be true right?

batmat: OK, there could also be 0. But is that really even a possibility? If so I guess yes a `logger.something` could be very useful, even more at that stage of the project.

services/src/services/update/update.class.js
```diff
@@ -0,0 +1,177 @@
...
+      $limit: 1,
+    };
+
+    /*
+     * By default, we want to take the latest Update Level, which is a:
+     *  ORDER BY createdAt DESC LIMIT 1
+     *
+     * For queries with levels, we want to take the _next_ Update Level, which
+     * is:
+     *  WHERE id  > ? ORDER BY createdAt ASC LIMIT 1
```
batmat: `AND tainted=false`, that is IIUC

services/ingest.yaml
```diff
@@ -1,49 +1,746 @@
...
-  aws:
+  docker-cloud:
     plugins:
-      - groupId: 'org.jenkins-ci.plugins'
-        artifactId: 'ec2'
-        url: 'https://updates.jenkins.io/download/plugins/ec2/1.39/ec2.hpi'
+      - groupId: io.jenkins.docker
+        artifactId: docker-plugin
+        url: >-
+          http://updates.jenkins-ci.org/download/plugins/docker-plugin/1.1.4/docker-plugin.hpi
```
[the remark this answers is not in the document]
rtyler: This file has been generated by a script from the update center

services/essentials.yaml
```diff
@@ -42,10 +42,22 @@ spec:
...
+      version: '2.73'
+
   environments:
     # environments get added to the other section when enabled
-    - name: aws
+    - name: docker-cloud
       plugins:
+        - groupId: 'io.jenkins.docker'
+          artifactId: 'docker-plugin'
+          version: '2.60.3'
```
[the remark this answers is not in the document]
rtyler: Yeah, presently unused because we're just grabbing values straight from the update-center unless otherwise hard-coded

services/src/services/update/update.class.js
rtyler: feathers-plus is, as far as I can tell, some third party project by [one person](https://github.com/feathers-plus) so I'm not sure I would call it "idiomatic".

We were already guarding this based on `internalOnly`, but the `undefined` completely removes the potential access to these methods and doesn't rely on a hook to do so. Additionally, it ensures that no stubbed out swagger documentation is generated for this

services/src/services/update/update.class.js
rtyler: A latestVersion of zero is a possibility for a fresh installation

rtyler:
I AM MERGING THIS BEFORE JENKINS CHANGES ITS MIND ZOMG

batmat:
^ was my review comments earlier today but I didn't finish it and didn't post the results. Sorry

rtyler:
@batmat No worries. I'll address some bits in subsequent pull request

distribution/client/src/client.js
```diff
@@ -87,9 +73,10 @@ class Client {
 
     this.errorTelemetry.setup();
 
-    setInterval( () => {
+    setInterval(() => {
       /* no-op to keep this process alive */
-    }, 10);
+      this.runUpdates();
+    }, 5000);
```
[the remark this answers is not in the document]
rtyler: I implemented it on this short timeframe and filed [JENKINS-51272](https://issues.jenkins-ci.org/browse/JENKINS-51272) to restore proper `Pusher` service so that, during integration tests, we can push a `check-updates` command to the client.

I'm starting on that ticket today, time permitting.

PR #107, opened 2018-06-15
Map compose runs to current user to avoid creating files owned by root

batmat:
That will avoid creating directories in the workspace with the root
user (current services/coverage).

I sometimes use `git clean -fdx -f` and this was failing because of the `services/coverage` directory.


batmat:
The "server" code depends a bit on file access permissions and so on, so as I did plan to just fix the `services/coverage` issue, I'm going to close this at least for now, this will need more work.

That work, BTW, might need to be done in a greater picture when we make, as I think we should, the services container run as non-root too like we did for the "distribution" client one some time ago.

PR #108, opened 2018-06-20
[JENKINS-51391] Adjust error telemetry API to JEP-308

mandie722:
[JENKINS-51391](https://issues.jenkins-ci.org/browse/JENKINS-51391)

I added a custom error handler to standardize the format of the error responses.

### DIFF
--- a/services/src/app.js
+++ b/services/src/app.js
@@ -28,6 +28,9 @@ const sequelizeSwagger = require('./sequelize-swagger');
 const settings = configuration();
 const app = express(feathers());
 
+// const SUCCESS = 'OK';
+const FAILURE = 'ERROR';
+
 // Load app configuration
 app.configure(settings);
 // Enable CORS, security, compression, favicon and body parsing
@@ -71,13 +74,19 @@ app.configure(channels);
 // Configure a middleware for 404s and the error handler
 app.use(express.notFound());
 
-/* Avoid cluttering the test logs with expected errors and exceptions */
-if (process.env.NODE_ENV != 'test') {
-  app.use(express.errorHandler({ logger }));
-}
-else {
-  app.use(express.errorHandler());
-}
+/* eslint-disable no-unused-vars */
+app.use((err, req, res, next) => { 
+  if (!err.statusCode) err.statusCode = err.code ? err.code : 500;
+  if (!err.message) err.message = 'Unexpected server error';
+  /* Avoid cluttering the test logs with expected errors and exceptions */
+  if (process.env.NODE_ENV != 'test') {
+    logger.error(err.stack);
+    logger.debug('statusCode: ' + err.statusCode + ' message: ' + err.message);
+  }
+  res.status(err.statusCode);
+  res.json({ status: FAILURE, message: err.message });
+});
+/* eslint-enable no-unused-vars */
 
 app.hooks(appHooks);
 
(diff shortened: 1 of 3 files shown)


services/src/app.js
```diff
@@ -28,6 +28,9 @@ const sequelizeSwagger = require('./sequelize-swagger');
 const settings = configuration();
 const app = express(feathers());
 
+// const SUCCESS = 'OK';
+const FAILURE = 'ERROR';
```
batmat: AFAICT, that is adding the response designed in the spec, but not checking the test actually receives this?

I tested locally using your fix, and adding the following inside the `.catch(err => {` block seems to do:

```javascript
assert.ok(err.error);
assert.ok(err.error.status);
assert.equal(err.error.status, 'ERROR');
```

services/src/app.js
```diff
@@ -71,13 +74,19 @@ app.configure(channels);
...
 
-/* Avoid cluttering the test logs with expected errors and exceptions */
-if (process.env.NODE_ENV != 'test') {
-  app.use(express.errorHandler({ logger }));
-}
-else {
-  app.use(express.errorHandler());
-}
+/* eslint-disable no-unused-vars */
+app.use((err, req, res, next) => {
```
batmat: AIUI, re-reading https://expressjs.com/en/guide/error-handling.html, we are missing the block:

```js
if (res.headersSent) {
  return next(err);
}
```

And if you agree, that would allow us to get rid of the eslint-disable above.

services/src/app.js
```diff
@@ -71,13 +74,19 @@ app.configure(channels);
...
-/* Avoid cluttering the test logs with expected errors and exceptions */
-if (process.env.NODE_ENV != 'test') {
-  app.use(express.errorHandler({ logger }));
-}
-else {
-  app.use(express.errorHandler());
-}
+/* eslint-disable no-unused-vars */
+app.use((err, req, res, next) => { 
+  if (!err.statusCode) err.statusCode = err.code ? err.code : 500;
```
batmat: It took me a bit of time to read this correctly, I think adding parenthesis would improve a bit readability (and also.

```js
err.statusCode = (err.code ? err.code : 500);
```

?

batmat (approved):
\o/

batmat:
Not merging yet mainly for the comment about `if (res.headersSent)`, but the rest LGTM and can be addressed in followups. The addition here already greatly improves the _statu quo_.

Congrats @mandie722 !

services/src/app.js
mandie722: Checking this actually will be my next task

services/src/app.js
mandie722: I was going back and forth on this, I can add this back in.

services/src/app.js
mandie722: Agree, fixing

batmat (approved):

batmat:
@mandie722 \o/. So FWIW, in that situation (currently "PR building" the PR):
image

i.e. when everything is approved, but the CI is still building, we wait for it to finish and be green, and then anyone can merge.

batmat:
I suspect https://github.com/jenkins-infra/evergreen/pull/109 will need a fix once it's merged. Going to first merge here Mandie's PR, then will address my https://github.com/jenkins-infra/evergreen/pull/109 following.

batmat:
It passed all the tests, but failed for unrelated reason because metrics 4.x plugin now depends on `variant`, and we use hardcoded latest as long as #105 is not in.

going to merge this, as anyway this is something also failing on master, and will fix.

PR #109, opened 2018-06-21
Enable more ESlint rules for codebase consistency

batmat:
Enabling a bit more rules so that the codestyle overall stays more consistent.

Also added the `fix-formatting` rule at the root `Makefile` to make fixing this when it happens even easier.
I didn't have to make a single fix manually, all were done using `make fix-formatting`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,3 +9,7 @@ lint:
 check:
 	$(MAKE) -C distribution $@
 	$(MAKE) -C services $@
+
+fix-formatting:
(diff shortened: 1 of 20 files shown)


rtyler (approved):

mandie722 (approved):
Seems valid to me, does improve visual consistency

batmat:
Unrelated failure for the same current reason as revealed in #110, will address there and merge here. Thanks for the reviews.

batmat:
FTR
image

PR #111, opened 2018-06-22
WIP Rtyler computed manifests 51752

batmat:
To test 105, but skipping `test_logs_are_not_under_jenkins_home`.


rtyler:
The test skip was incorporated into #105, thanks @batmat !

PR #112, opened 2018-06-22
JENKINS-52143 - add validation on the JSON responses

mandie722:
Just added in some assertions around the JSON body on the responses for success and failure.

### DIFF
--- a/services/acceptance/services/errortelemetry.test.js
+++ b/services/acceptance/services/errortelemetry.test.js
@@ -53,6 +53,8 @@ describe('Error Telemetry service acceptance tests', () => {
           .then(res => assert.fail(res))
           .catch(err => {
             assert.equal(err.statusCode, 400);
+            assert.equal(err.error.status, 'ERROR', 'response.status should be ERROR but got: ' + err.error.status);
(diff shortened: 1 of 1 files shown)


services/acceptance/services/errortelemetry.test.js
```diff
@@ -119,7 +124,10 @@ describe('Error Telemetry service acceptance tests', () => {
           json: true,
           body: emptyLog
         })
-          .then(res => assert.ok(res))
+          .then(res => {
+            assert.ok(res);
+            assert.equal(res.status, 'OK', 'response.status should be OK but got: ' + res.status);
```
rtyler: One of the things @batmat taught me was about [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in JavaScript (as of ECMAScript 6) which you may find helpful.

mandie722: Is templating preferred over string concatenation?

rtyler: I would recommend it, it definitely makes life easier IMHO.

There's no way to enforce that convention that I have found with ESLint however.

PR #116, opened 2018-06-25
Update to latest updates

batmat:
No manual change. Only ran `make -C services ingest-update-center`


batmat:
Merging as this succeeded, and this is going to create a bunch of conflicts of big bumps later if we wait more. Also as I'm adding more things for AWS, I would like it to be as up to date as possible to limit updates there to aws ec2 related things.

PR #120, opened 2018-06-26
[JENKINS-52122] check essentials and metrics plugins are installed

batmat:
https://issues.jenkins-ci.org/browse/JENKINS-52122

Goal is to avoid losing time each time one changes the plugin list, and might end up removing those Essentials-required plugins unintendedly.


batmat:
OK, build success after close/reopen. Didn't change code since Tyler's review, hence merging.

PR #121, opened 2018-06-27
[JENKINS-52163] update acceptance tests to use expect

mandie722:
[JENKINS-52163](https://issues.jenkins-ci.org/browse/JENKINS-52163)

Update acceptance tests to use expect where possible.  There's no simple expect equivalent to assert.fail.

Added a hook to the update create path to fix a test so that it fails as expected.

### DIFF
--- a/services/acceptance/services/registration.test.js
+++ b/services/acceptance/services/registration.test.js
@@ -12,13 +12,17 @@ describe('Registration service acceptance tests', () => {
         url: h.getUrl('/registration'),
         method: 'POST',
         json: true,
+        resolveWithFullResponse: true,
         body: {
           pubKey: 'pretend-pubkey',
           curve: 'secp256k1'
         }
       })
-        .then(res => assert.ok(res.uuid, 'Missing a uuid'))
-        .catch(res => assert.equal(res.statusCode, 201));
+        .then(res => {
+          expect(res.statusCode).toEqual(201);
+          expect(res.body.uuid).toBeTruthy();
+        })
+        .catch((err) => assert.fail(err));
     });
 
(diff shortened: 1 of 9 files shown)


services/src/services/update/update.hooks.js
```diff
@@ -18,6 +19,19 @@ class UpdateHooks {
     return context;
   }
 
+  checkUpdateFormat(hook) {
+    if (!hook) {
+      throw new errors.BadRequest('No hook at all?');
```
rtyler: This check is unnecessary, Feathers hooks are always guaranteed to be called with a hook `context` with a [specified format](https://docs.feathersjs.com/api/hooks.html#hook-context)

services/acceptance/services/registration.test.js
```diff
@@ -12,13 +12,17 @@ describe('Registration service acceptance tests', () => {
...
           curve: 'secp256k1'
         }
       })
-        .then(res => assert.ok(res.uuid, 'Missing a uuid'))
-        .catch(res => assert.equal(res.statusCode, 201));
+        .then(res => {
+          expect(res.statusCode).toEqual(201);
+          expect(res.body.uuid).toBeTruthy();
+        })
+        .catch((err) => assert.fail(err));
```
batmat: I guess we cannot indeed use something like:

```js
expect(request({
        url: h.getUrl('/registration'),
        method: 'POST',
        json: true,
        body: {
          pubKey: 'pretend-pubkey'
        }
      }).toThrow(BLAH); 
```

Because there seems to be no way to express the check instead of `BLAH` that `err.statusCode == 400`

PR #122, opened 2018-06-27
Attempt to refactor the error telemetry code to avoid weird failures

rtyler:


batmat:
Though I cannot check in trusted infra, I think this does not fix the issue, I'm still seeing it on an EC2 instance. Though no idea yet what is causing it.

distribution/client/test/downloader.test.js
```diff
@@ -1,11 +1,11 @@
-'use strict';
+jest.mock('fs');
 
 const Downloader = require('../src/lib/downloader');
 
 describe('the Downloader class', () => {
   describe('download()', () => {
     let item = 'https://jenkins.io';
-    let dir  = '/tmp/jest';
+    let dir  = '/tmp';
```
batmat: I think there's an issue here, since it will try to write a file into a directory. So it could be linked to the error `EISDIR: illegal operation on a directory`, though I do not understand the link with `open '/evergreen/jenkins/home'`

PR #123, opened 2018-06-28
[JENKINS-52234] Make sure directory exists

batmat:


batmat:
Testing this in parallel through a manual `make check` on an EC2 instance where I was seeing JENKINS-52234. Will merge if both here, and there on that instance succed.

batmat:
OK, merging as this succeeded on the EC2 instance and here, and I'd like to see if this ends up working in the "trusted" CI and we get the Docker images deployed.

PR #125, opened 2018-06-28
Name test runs folders with flavor and test file name to ease diagnostics

batmat:
In CI and locally, it can be very time consuming to know which context a given log file belongs to.

Here is what this change looks like in CI:
https://ci.jenkins.io/blue/organizations/jenkins/Infra%2Fevergreen/detail/PR-125/18/artifacts

image


batmat:
Will merge when PR build is green.

mandie722 (approved):

PR #131, opened 2018-07-10
[JENKINS-52455] Remove port 50000

batmat:
[JENKINS-52455](https://issues.jenkins-ci.org/browse/JENKINS-52455)

Fixing issue noticed by @dcoraboeuf on https://github.com/jenkinsci/jep/pull/140#discussion_r201321330 Thanks Damien!

### DIFF
--- a/distribution/environments/aws-ec2-cloud/CloudFormation/cloudformation-template.json
+++ b/distribution/environments/aws-ec2-cloud/CloudFormation/cloudformation-template.json
@@ -53,7 +53,7 @@
               "export DOCKER_IMAGE=jenkins/evergreen:aws-ec2-cloud\n",
               "echo \"", { "Ref": "PrivateKey"}, "\" > ssh-agents-private-key\n",
               "docker pull $DOCKER_IMAGE\n",
-              "docker run -d -p 8080:8080 -p 50000:50000 ",
+              "docker run -d -p 8080:8080 ",
               " --name jenkins-essentials",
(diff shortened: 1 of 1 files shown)


distribution/environments/aws-ec2-cloud/CloudFormation/cloudformation-template.json
```diff
@@ -53,7 +53,7 @@
               "export DOCKER_IMAGE=jenkins/evergreen:aws-ec2-cloud\n",
               "echo \"", { "Ref": "PrivateKey"}, "\" > ssh-agents-private-key\n",
               "docker pull $DOCKER_IMAGE\n",
-              "docker run -d -p 8080:8080 -p 50000:50000 ",
```
rtyler: This will prevent any JNLP-based agents from being launchable from Jenkins Essentials, such as Windows agents.

I don't think that's a desirable outcome, regardless of how undesirable Windows agents may be :smiley_cat: 


I think restricting 50000 to the security group makes more sense than removing this entirely

rtyler (changes requested):
JNLP agents are a still a thing :slightly_frowning_face:

batmat:
OK, going to close this finally:

* the port is anyway currently **not** exposed/open from Jenkins
* Leaving it available as is currently will make it available for Essentials users to enable a fixed port so that they can connect agents if they wish to. If we close it at both Docker ports, and AWS security groups levels, and they want to enable it, well they are going to have a very hard time.

PR #133, opened 2018-07-10
Jenkins 51735 sentry evaluation

mandie722:
https://issues.jenkins-ci.org/browse/JENKINS-51735

Added in methods to send output to sentry.io from the error telemetry apis.

### DIFF
--- a/services/src/services/errorTelemetry/errorTelemetry.class.js
+++ b/services/src/services/errorTelemetry/errorTelemetry.class.js
@@ -2,8 +2,7 @@ const fs     = require('fs');
 const logger = require('winston');
 const path   = require('path');
 const mkdirp = require('mkdirp');
-const Raven  = require('raven');
-
+const sentry = require('../../lib/sentry');
 
 const DEFAULT_ERROR_LOGGING_FILE = '/srv/evergreen/error-logging.json';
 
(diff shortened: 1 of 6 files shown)


batmat:
Error looks legit. Closing this to avoid it to rebuild every 6 hours @mandie722, just reopen it when you're back and think you fixed the issue. Thanks

services/src/services/errorTelemetry/errorTelemetry.class.js
```diff
@@ -27,9 +27,9 @@ class ErrorTelemetryService {
       return Promise.reject({status:'KO'});
     }
 
-    // FIXME: TBD where, what and how to actually send data
     const toWrite = `${new Date()} => ${JSON.stringify(data)}\n\n`;
```
batmat: I do not think that comment should be removed yet, or the format should be evolved a bit. As it is now, it's sending a somewhat still raw: "date (without defined format) => the json of the log", which I fear is not going to be greatly parsed by Sentry is it? Anyway, I guess we should clean up that too.

services/src/services/errorTelemetry/errorTelemetry.class.js
```diff
@@ -2,7 +2,7 @@ const fs     = require('fs');
 const logger = require('winston');
 const path   = require('path');
 const mkdirp = require('mkdirp');
-
+const sentry = require('../../lib/sentry');
```
batmat: The CI error seems to come from this. I seem to understand the `raven` module got imported above, but then the code references a manually (?) added lib from two directories above?

I seem to understand that the line above should just be replaced by 

```js
const sentry = require('raven');
```

And then it would work?

PR #134, opened 2018-07-10
Refactor some tests to hopefully appease the trusted.ci gods

rtyler:
There's some oddities when playing with `/tmp` and mocked file systems in trusted.ci, but of course I cannot reproduce these locally or anywhere else. So I'm hoping this helps, even maybe a wee bit.

### DIFF
--- a/distribution/client/test/downloader.test.js
+++ b/distribution/client/test/downloader.test.js
@@ -13,13 +13,13 @@ describe('the Downloader class', () => {
     });
 
     it('should return promise', () => {
-      let response = Downloader.download('https://jenkins.io/index.html', dir);
+      let response = Downloader.download(`${item}/index.html`, dir);
       expect(Promise.resolve(response)).toBe(response);
     });
 
     it('should fail on url without final basename-ish path', () => {
       expect(() => {
-        Promise.resolve(Downloader.download('https://jenkins.io/', dir));
+        Promise.resolve(Downloader.download(`${item}/`, dir));
       }
       ).toThrow();
     });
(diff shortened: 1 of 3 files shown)


distribution/client/test/downloader.test.js
```diff
@@ -1,24 +1,25 @@
...
+jest.mock('fs');
 
 const Downloader = require('../src/lib/downloader');
 const mkdirp     = require('mkdirp');
 
 describe('the Downloader class', () => {
   describe('download()', () => {
-    let dir  = '/tmp/jest';
+    let item = 'https://jenkins.io';
+    let dir  = '/tmp';
```
batmat: why not `const`?

distribution/client/test/error-telemetry.test.js
```diff
@@ -2,55 +2,56 @@ jest.mock('fs');
...
     });
   });
 
   describe('setup() call', () => {
+    const errorTelemetryService = new ErrorTelemetry();
 
-    // FIXME: only hackish, the end goal is definitely not to forward to another file
-    it('writing to essentials logging file should forward to another', done => {
+    let logsDir = '/evergreen/jenkins/war/logs';
+    let logFile = path.join(logsDir, 'essentials.log.0');
```
batmat: same here with `const`?

distribution/client/test/error-telemetry.test.js
```diff
@@ -2,55 +2,56 @@ jest.mock('fs');
...
 
-      // Given: the file is watched
```
batmat: NIT: you've removed the `Given` of the `given, when, then` suite.

PR #135, opened 2018-07-11
Build just once a day

batmat:
The current settings also post on PRs every 6 hours a failure message
when they are failing. We do not really need such a high frequency IMO,
~once a day on master was the main idea for this addition to catch
issues earlier.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ pipeline {
 
     triggers {
         pollSCM('H * * * *')
-        cron('H */6 * * *')
+        cron('H H * * *')
     }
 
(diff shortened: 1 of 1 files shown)


batmat:
```
test_npm_5_plus

ASSERT:expected:<5.> but was:<OC>

shunit2:ERROR test_npm_5_plus() returned non-zero return code.

test_jep_307
```

And, in the logs from the container:

```
2018-07-11 07:05:55,176 INFO spawnerr: can't find command '/usr/bin/npm'
```

OK, `npm` is absent, everything is normal... 
Wait, what?

PR #138, opened 2018-07-11
[JENKINS-51735] Sentry evaluation

mandie722:
https://issues.jenkins-ci.org/browse/JENKINS-51735

Added in methods to send output to sentry.io from the error telemetry apis.

### DIFF
--- a/services/Makefile
+++ b/services/Makefile
@@ -2,6 +2,9 @@ PATH:=./node_modules/.bin:../tools:$(PATH)
 COMPOSE:=../tools/compose
 IMAGE_NAME:=jenkinsciinfra/evergreen-backend
 DB_DUMP=initdb.d/db.sql
+ifndef SENTRY_URL
+  SENTRY_URL='http://sentry.example.com'
+endif
 
 all: check docs container
 
--- a/services/Makefile
+++ b/services/Makefile
@@ -3,7 +3,7 @@ COMPOSE:=../tools/compose
 IMAGE_NAME:=jenkinsciinfra/evergreen-backend
 DB_DUMP=initdb.d/db.sql
 ifndef SENTRY_URL
-  SENTRY_URL='http://sentry.example.com'
+  SENTRY_URL='https://publickey@sentry.io/projectid'
 endif
 
 all: check docs container
--- a/services/src/libs/sentry.js
+++ b/services/src/libs/sentry.js
@@ -1,0 +1,46 @@
+/*
+ * The sentry module is responsible for passing messages along to
+ * the sentry server.
+ */
+
+const logger    = require('winston');
+const Raven     = require('raven');
+
+
+class Sentry {
+  static initialize(sentryUrl) {
+    if (!sentryUrl) {
+      logger.error('No sentry url defined.');
+      return;
+    }
+    Raven.config(sentryUrl).install();
+  }
+
+  static sendOutput(data) {
+    if (!data) {
+      logger.error('Missing data.');
+      return;
+    }
+    if (data.log.exception) {
+      Raven.captureException(new Error(data.log.message), {
+        level: data.log.level.toLowerCase(),
+        logger: data.log.name,
+        extra: {
+          uuid: data.uuid,
+          source: data.log
+        }
+      });
+    } else {
+      Raven.captureMessage(data.log.message, {
+        level: data.log.level.toLowerCase(),
+        logger: data.log.name,
+        extra: {
+          uuid: data.uuid,
+          source: data.log
+        }
+      });
+    }
+  }
+}
+
+module.exports = Sentry;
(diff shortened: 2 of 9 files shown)


services/src/services/errorTelemetry/errorTelemetry.class.js
```diff
@@ -27,9 +27,9 @@ class ErrorTelemetryService {
       return Promise.reject({status:'KO'});
     }
 
-    // FIXME: TBD where, what and how to actually send data
     const toWrite = `${new Date()} => ${JSON.stringify(data)}\n\n`;
     fs.appendFileSync(this.loggingFile, toWrite);
```
rtyler: The file appending can certainly be removed with the Sentry integration

What is going to have to change, and I'm curious what solution you and @batmat come up with, is how we would verify from our smoke test whether the Sentry logging is working correctly. While I don't think that we should be sending exceptions to a live Sentry instance, some level of validation that something _should_ have been passed to Sentry from a running smoke test is important IMO.

services/src/libs/sentry.js
```diff
@@ -0,0 +1,46 @@
...
+class Sentry {
+  static initialize(sentryUrl) {
+    if (!sentryUrl) {
+      logger.error('No sentry url defined.');
+      return;
+    }
+    Raven.config(sentryUrl).install();
+  }
+
+  static sendOutput(data) {
```
rtyler: What is the expected behavior if `sendOutput` is invoked without `initiatialize` having run successfully?

services/src/libs/sentry.js
```diff
@@ -0,0 +1,46 @@
...
+    if (!data) {
+      logger.error('Missing data.');
+      return;
+    }
+    if (data.log.exception) {
+      Raven.captureException(new Error(data.log.message), {
+        level: data.log.level.toLowerCase(),
```
rtyler: The object passed to both `captureMessage` and `captureException` looks identical to me, perhaps that can be refactored.

services/src/libs/sentry.js
```diff
@@ -0,0 +1,46 @@
+/*
+ * The sentry module is responsible for passing messages along to
+ * the sentry server.
+ */
+
+const logger    = require('winston');
+const Raven     = require('raven');
+
+
+class Sentry {
```
rtyler: Please add gratuitous JSDoc to the class and methods :smiley_cat:

services/src/services/errorTelemetry/errorTelemetry.class.js
mandie722: I've been looking into a mocking option that will allow us to unit test the sentry class, I'll create a ticket for that to make sure it gets done.

services/src/libs/sentry.js
mandie722: it looks different in sentry which is why it's called differently.

services/src/services/errorTelemetry/errorTelemetry.class.js
mandie722: Also, I wouldn't remove the file appending until we're 100% on a final end solution.  Even then, it's useful to have a backup of what data has been received since we don't track it any other way.

services/src/libs/sentry.js
mandie722: if initialize is not called, then the entire module throws errors.  It's a part of the application startup.

batmat:
@mandie722 Coverage threshold failure FYI

batmat:
@mandie722 just merged with `master` to get the coverage failure locally too, thanks to https://github.com/jenkins-infra/evergreen/pull/141. Then I added a very small test to fix the coverage issue. Hopefully, at least the pipeline will now fail a bit later and save you a bit of iteration when you're back :-). Just remember to update your clone with the changes I pushed to your branch :).

services/Makefile
```diff
@@ -2,6 +2,9 @@ NODE:=../tools/node
 COMPOSE:=../tools/compose
 IMAGE_NAME:=jenkinsciinfra/evergreen-backend
 DB_DUMP=initdb.d/db.sql
+ifndef SENTRY_URL
+  SENTRY_URL='https://publickey@sentry.io/projectid'
```
rtyler: I think it would be best to incorporate this into `config/default.json` and then allow an environment override for production environments like with `EVERGREEN_JWT_SECRET`

services/src/services/errorTelemetry/errorTelemetry.class.js
rtyler: The file appending is not going to work in a production environment since we'll be deployed as a stateless set of containers in Kubernetes. We won't be able to sensibly retrieve the contents of the file system, but if we miss an errant exception or two, I won't be terribly sad.

PR #139, opened 2018-07-12
Fully configure artifact storage in S3 using CasC

batmat:
Updating to latest configuration-as-code plugin too to allow this kind of config + be more picky about bad attribute (e.g. zone on the ec2-cloud.yaml that was misplaced and unused until now).

Tested manually :sob:. Really need to give more thoughts on how to automate tests, and do it.


rtyler (approved):

carlossg:
LGTM

PR #144, opened 2018-07-16
[JENKINS-52304] Adjust artifact-manager-s3 configuration acc. to refactoring in aws-global-configuration

jglick:
Amends #139 to take into account https://github.com/jenkinsci/aws-global-configuration-plugin/pull/4 + https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/72. @batmat @carlossg @kuisathaverat

### DIFF
--- a/distribution/environments/aws-ec2-cloud/config/as-code/artifact-manager-s3.yaml
+++ b/distribution/environments/aws-ec2-cloud/config/as-code/artifact-manager-s3.yaml
@@ -1,14 +1,14 @@
 ---
-# TODO: file JIRAs to make this classifed (and for other used plugins too)
 unclassified:
-  s3blobstoreconfig:
-    # TODO: create through CloudFormation
-    container: "${ARTIFACT_MANAGER_S3_BUCKET_NAME}"
-    # FIXME: not a sensible default for everyone
-    region: "us-east-1"
-    prefix: "jenkins_data/"
-
   artifactManager:
     artifactManagerFactories:
-      - JCloudsArtifactManagerFactory:
-          provider: s3blobstore
+      - jclouds:
+          provider: s3
+aws:
+  awsCredentials:
+    # FIXME: not a sensible default for everyone
+    region: "us-east-1"
+  s3:
+    # TODO: create through CloudFormation
+    container: "${ARTIFACT_MANAGER_S3_BUCKET_NAME}"
+    prefix: "jenkins_data/"
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -12,6 +12,7 @@ spec:
   plugins:
     - groupId: 'org.jenkins-ci.plugins'
       artifactId: 'configuration-as-code'
+      # TODO overridden
       version: '0.1-alpha'
 
     - groupId: 'io.jenkins.blueocean'
@@ -49,10 +50,13 @@ spec:
       artifactId: 'jdk-tool'
       version: '2.73'
 
-# FIXME: delete the two deps below for AWS once JENKINS-51766 is done
+# FIXME: delete the deps below for AWS once JENKINS-51766 is done
     - groupId: 'org.jenkins-ci.plugins'
       artifactId: 'ec2'
       version: '1.39'
+    - groupId: 'io.jenkins.plugins'
+      artifactId: 'aws-global-configuration'
+      version: '1.0'
     - groupId: 'io.jenkins.plugins'
       artifactId: 'artifact-manager-s3'
       version: '1.0'
@@ -72,7 +76,12 @@ spec:
           artifactId: 'ec2'
           version: '1.39'
         - groupId: 'io.jenkins.plugins'
+          artifactId: 'aws-global-configuration'
+          # TODO overridden
+          version: '1.0'
+        - groupId: 'io.jenkins.plugins'
           artifactId: 'artifact-manager-s3'
+          # TODO overridden
           version: '1.0'
 
 # the realized BoM after refs are resolved
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -59,7 +59,7 @@ spec:
       version: '1.0'
     - groupId: 'io.jenkins.plugins'
       artifactId: 'artifact-manager-s3'
-      version: '1.0'
+      version: '1.1'
 # END FIXME shim to remove for JENKINS-51766
 
 
@@ -77,12 +77,10 @@ spec:
           version: '1.39'
         - groupId: 'io.jenkins.plugins'
           artifactId: 'aws-global-configuration'
-          # TODO overridden
           version: '1.0'
         - groupId: 'io.jenkins.plugins'
           artifactId: 'artifact-manager-s3'
-          # TODO overridden
-          version: '1.0'
+          version: '1.1'
 
 # the realized BoM after refs are resolved
 # status:
--- a/services/ingest-update-center
+++ b/services/ingest-update-center
@@ -18,7 +18,7 @@ let ingest = {
   },
   plugins: [
     /*
-     * Hard-coding configuration as code until it hits the main update center
+     * Hard-coding some plugins until they hit the main update center
      */
     {
       groupId: 'org.jenkins-ci.plugins',
@@ -29,15 +29,33 @@ let ingest = {
         signature: '4faefcc46299557efaa1c896835a2413ad0646201e40938b4207dbd4cfecf33303cf5f28fd1f53cb71146c45d12c7fa34d2d87d312acf9fd7964c08506d2278d',
       },
     },
-     {
-       groupId: 'io.jenkins.plugins',
-       artifactId: 'essentials',
-       url: 'http://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/essentials/0.3-rc20.a3f76cdd1348/essentials-0.3-rc20.a3f76cdd1348.hpi',
-       checksum: {
-         type: 'sha512',
-         signature: 'e1cdf271eabed428c2198c4f7a49c5dea7fc8850987bcf63b87b74936a3cc426896a4b7c9c974f0fe3abd61071ea5a2238538af9f9266f2d37696856809d697a',
-       },
-     },
+    {
+      groupId: 'io.jenkins.plugins',
+      artifactId: 'aws-global-configuration',
+      url: 'https://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/aws-global-configuration/1.0-rc20.ffbe7e5de068/aws-global-configuration-1.0-rc20.ffbe7e5de068.hpi',
+      checksum: {
+        type: 'sha512',
+        signature: 'c78067f7f96668aa9798b7fc735b77dfaa208f4499238d42079ba6a4c9eabbc43adf6b58b9e5cb1dd143b17659f33d23945cda86d09423e17c4b9dad682c5cef',
+      },
+    },
+    {
+      groupId: 'io.jenkins.plugins',
+      artifactId: 'artifact-manager-s3',
+      url: 'https://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/artifact-manager-s3/1.1-rc77.15d7dd937b42/artifact-manager-s3-1.1-rc77.15d7dd937b42.hpi',
+      checksum: {
+        type: 'sha512',
+        signature: '547b6a605aae7585f1c4d0995618735942ce3936bf4560bd11717b7c0c08876d30933a755a596f5be6e7d0834b2e3587fa55913fe3233a6d28857976435aa18d',
+      },
+    },
+    {
+      groupId: 'io.jenkins.plugins',
+      artifactId: 'essentials',
+      url: 'http://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/essentials/0.3-rc20.a3f76cdd1348/essentials-0.3-rc20.a3f76cdd1348.hpi',
+      checksum: {
+        type: 'sha512',
+        signature: 'e1cdf271eabed428c2198c4f7a49c5dea7fc8850987bcf63b87b74936a3cc426896a4b7c9c974f0fe3abd61071ea5a2238538af9f9266f2d37696856809d697a',
+      },
+    },
   ],
   environments: {
   },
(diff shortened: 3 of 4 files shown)


distribution/environments/aws-ec2-cloud/config/as-code/artifact-manager-s3.yaml
```diff
@@ -1,14 +1,14 @@
...
 unclassified:
-  s3blobstoreconfig:
-    # TODO: create through CloudFormation
-    container: "${ARTIFACT_MANAGER_S3_BUCKET_NAME}"
+  artifactManager:
+    artifactManagerFactories:
+      - jclouds:
+          provider: s3
+aws:
+  awsCredentials:
```
jglick: Note that while https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/72 tests this format, I did not attempt to test this actual file patch, since I have no idea how to do so—acc. to discussion with @batmat there appears to be no equivalent to

```sh
make ENV=aws try-everything
```

which would use (say) Docker Compose to start up a local copy of the backend and the client so you can go to `http://localhost:8080/` and poke around.

Also it seems that (acc. to [JENKINS-52581](https://issues.jenkins-ci.org/browse/JENKINS-52581)) `Jenkinsfile` currently runs no tests which would even verify that this file applies without error.

services/essentials.yaml
```diff
@@ -12,6 +12,7 @@ spec:
   plugins:
     - groupId: 'org.jenkins-ci.plugins'
       artifactId: 'configuration-as-code'
+      # TODO overridden
```
jglick: Reminder that the version here is ignored, or is supposed to be.

services/ingest-update-center
```diff
@@ -29,15 +29,33 @@ let ingest = {
...
+      artifactId: 'artifact-manager-s3',
+      url: 'https://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/artifact-manager-s3/1.1-rc77.15d7dd937b42/artifact-manager-s3-1.1-rc77.15d7dd937b42.hpi',
+      checksum: {
+        type: 'sha512',
+        signature: '547b6a605aae7585f1c4d0995618735942ce3936bf4560bd11717b7c0c08876d30933a755a596f5be6e7d0834b2e3587fa55913fe3233a6d28857976435aa18d',
+      },
+    },
+    {
+      groupId: 'io.jenkins.plugins',
+      artifactId: 'essentials',
```
jglick: just correcting indentation

services/ingest-update-center
```diff
@@ -29,15 +29,33 @@ let ingest = {
...
+      url: 'http://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/essentials/0.3-rc20.a3f76cdd1348/essentials-0.3-rc20.a3f76cdd1348.hpi',
```
jglick: `http` or `https` preferred? Inconsistent between `essentials` and `configuration-as-code`.

services/ingest-update-center
```diff
@@ -48,12 +66,18 @@ let ingest = {
...
+  let dependencies = plugin ? plugin.dependencies : undefined;
+  if (dependencies) {
+    dependencies.forEach((dependency) => {
+      if (!dependency.optional) {
+        depends.push(dependency.name);
+        resolveFor(dependency.name, graph).forEach(d => depends.push(d));
+      }
+    });
+  } else {
+    logger.info(`Skipping dependency processing for ${pluginName}, you had better make sure all deps are already in essentials.yaml`);
```
jglick: Hack alert. As noted in Gitter, this is risky. Really the whole script needs to be rewritten: rather than load `update-center.json`, which is useless for Incrementals, it should directly download `*.hpi` and examine manifests.

services/ingest-update-center
```diff
@@ -62,7 +86,15 @@ let resolveFor = function(pluginName, graph) {
  * Convert a given plugin name to the ingest.yaml representation of it
  */
 let pluginToRecord = function(plugin, graph) {
-  let gav = graph.plugins[plugin].gav;
+  let override = ingest.plugins.find(record => record.artifactId === plugin);
+  if (override) {
+    return override;
```
jglick: Another hack.

services/ingest-update-center
```diff
@@ -62,7 +86,15 @@ let resolveFor = function(pluginName, graph) {
...
+  }
+  let pluginData = graph.plugins[plugin];
+  if (!pluginData) {
+    logger.info(`No plugin data for ${plugin}`);
```
jglick: Just an explanation before the `TypeError`.

services/ingest-update-center
```diff
@@ -88,7 +120,11 @@ essentials.spec.plugins.forEach((plugin) => {
 });
 plugins = [...new Set(plugins)];
 plugins.forEach((plugin) => {
-  ingest.plugins.push(pluginToRecord(plugin, uc));
+  if (ingest.plugins.find(record => record.artifactId === plugin)) {
+    logger.info(`Using override for ${plugin}`);
```
jglick: Another hack.

services/ingest.json
```diff
@@ -1,5 +1,5 @@
 {
-  "timestamp": "2018-07-12T10:22:35.628Z",
+  "timestamp": "2018-07-16T14:39:03.906Z",
```
jglick: This ought to be injected dynamically by the backend service IMO.

services/ingest.json
```diff
@@ -17,6 +17,24 @@
...
+      "artifactId": "aws-global-configuration",
+      "url": "https://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/aws-global-configuration/1.0-rc20.ffbe7e5de068/aws-global-configuration-1.0-rc20.ffbe7e5de068.hpi",
+      "checksum": {
+        "type": "sha512",
+        "signature": "c78067f7f96668aa9798b7fc735b77dfaa208f4499238d42079ba6a4c9eabbc43adf6b58b9e5cb1dd143b17659f33d23945cda86d09423e17c4b9dad682c5cef"
+      }
+    },
+    {
+      "groupId": "io.jenkins.plugins",
+      "artifactId": "artifact-manager-s3",
```
jglick: Apparently `ingest-update-center` neglects to sort the plugin list, so diffs will be hard to verify.

services/ingest.json
```diff
@@ -29,55 +47,55 @@
     {
       "groupId": "io.jenkins.blueocean",
       "artifactId": "blueocean",
-      "url": "http://updates.jenkins-ci.org/download/plugins/blueocean/1.6.2/blueocean.hpi",
+      "url": "http://updates.jenkins-ci.org/download/plugins/blueocean/1.7.0/blueocean.hpi",
```
jglick: As noted in Gitter, so long as transitive dependencies are omitted from `essentials.yaml` and their versions picked up from the UC, your manifest will be **nondeterministic**, which is terrible. Best to force the editor of `essentials.yaml` to specify every plugin to be included and its version, and fail the `make` target if there is any inconsistency (missing plugin, plugin too old, plugin incompatible with Jenkins core, etc.). Failing that, at least use the lowest upper bound of the dependency versions requested in the graph, instead of looking in `update-center.json`, so that the resulting version is deterministic.

services/ingest.json
```diff
@@ -17,6 +17,24 @@
         "signature": "4faefcc46299557efaa1c896835a2413ad0646201e40938b4207dbd4cfecf33303cf5f28fd1f53cb71146c45d12c7fa34d2d87d312acf9fd7964c08506d2278d"
       }
     },
+    {
+      "groupId": "io.jenkins.plugins",
+      "artifactId": "aws-global-configuration",
+      "url": "https://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/aws-global-configuration/1.0-rc20.ffbe7e5de068/aws-global-configuration-1.0-rc20.ffbe7e5de068.hpi",
+      "checksum": {
+        "type": "sha512",
+        "signature": "c78067f7f96668aa9798b7fc735b77dfaa208f4499238d42079ba6a4c9eabbc43adf6b58b9e5cb1dd143b17659f33d23945cda86d09423e17c4b9dad682c5cef"
```
jglick: This is the wrong format I guess? Why is the SHA-512 in `ingest-update-center` expressed in hex while in `ingest.json` is it Base64 of what I presume is binary? Yet the mismatch was already there in `configuration-as-code` and `essentials`. So is the client not actually checking the checksum, or what??

distribution/environments/aws-ec2-cloud/config/as-code/artifact-manager-s3.yaml
```diff
@@ -1,14 +1,14 @@
...
     # FIXME: not a sensible default for everyone
     region: "us-east-1"
+  s3:
+    # TODO: create through CloudFormation
+    container: "${ARTIFACT_MANAGER_S3_BUCKET_NAME}"
     prefix: "jenkins_data/"
```
jglick: BTW I would suggest the prefix be empty; not sure why we would want to specify one. But up to @carlossg to comment.

jglick:
Some issue in `test_metrics_health_check`. I see

```
/usr/local/bin/docker: Cannot connect to the Docker daemon at tcp://localhost:2375. Is the docker daemon running?.
```

but I am not sure if that is somehow related to my patch. Also

```
./shunit2/shunit2: line 950: [: skipped=5: integer expression expected
```

whatever that means.

distribution/environments/aws-ec2-cloud/config/as-code/artifact-manager-s3.yaml
carlossg: I think it is useful to allow putting other data in the same bucket. I would have called it `artifacts/` though

jglick: There is already a subdirectory for artifacts, as well as one for stashes. Currently these are under the build prefix. Maybe we would want to call the default prefix `job/` or `builds/` or something like that then? Or did we want a _top-level` subdirectory for artifacts and one for stashes, to make it easier to apply S3 retention rules—is it possible to do these with glob-type matches or is that unreliable?

batmat:
@jglick Seems to be a legit issue. See archived test run logs, like https://ci.jenkins.io/job/Infra/job/evergreen/job/PR-144/2/artifact/distribution/build/tests-run-2018-07-17_04-46-13-docker-cloud-tests.sh/distribution_instance_1.log

```
[INFO][2018-07-17 04:45:06] Listed all plugins (from jenkins.InitReactorRunner$1 onAttained)
[SEVERE][2018-07-17 04:45:06] Failed Loading plugin Artifact Manager on S3 plugin v1.1-rc77.15d7dd937b42 (artifact-manager-s3) (from jenkins.InitReactorRunner$1 onTaskFailed)
java.io.IOException: Artifact Manager on S3 plugin v1.1-rc77.15d7dd937b42 failed to load.
 - Pipeline: API v2.28 is older than required. To fix, install v2.29-rc313.bb1f9f990cbc or later.
 - AWS Global Configuration Plugin v1.0-rc20.ffbe7e5de068 is older than required. To fix, install v1.0-rc20.980a9cfa0b14 or later.
	at hudson.PluginWrapper.resolvePluginDependencies(PluginWrapper.java:655)
	at hudson.PluginManager$2$1$1.run(PluginManager.java:516)
	at org.jvnet.hudson.reactor.TaskGraphBuilder$TaskImpl.run(TaskGraphBuilder.java:169)
	at org.jvnet.hudson.reactor.Reactor.runTask(Reactor.java:296)
	at jenkins.model.Jenkins$5.runTask(Jenkins.java:1068)
	at org.jvnet.hudson.reactor.Reactor$2.run(Reactor.java:214)
	at org.jvnet.hudson.reactor.Reactor$Node.run(Reactor.java:117)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)

```

Did `make container-check` run successfully for you locally?

distribution/environments/aws-ec2-cloud/config/as-code/artifact-manager-s3.yaml
batmat: So, I was unclear I think. We do not have a way currently to test the specific image built for AWS, that is indeed the gist of [JENKINS-52581](https://issues.jenkins-ci.org/browse/JENKINS-52581). But by running either `make check` from the repo root, or the more specific `make container-check` under the `distribution` directory, you would already run a bunch of tests. Seems one is already failing, see https://github.com/jenkins-infra/evergreen/pull/144#issuecomment-405567940 below.

batmat: Build running in https://github.com/jenkins-infra/evergreen/pull/146 @jglick if this succeeds, as I think it should, you should be able to use it locally to check at least the server starts for instance (which is not the case currently).

jglick:
> See archived test run logs

How was I supposed to know to look for that? The build console did not include that (at least not at the end where I looked), and there were no JUnit-format reports.

> Did `make container-check` run successfully for you locally?

I did not try any `make` targets locally since I did not know what targets to select. That is what I was asking about being able to verify behavior of the `aws` environment at least interactively.

At this point it is too late to use Essentials to validate incremental changes: all of the relevant code is now in MRP versions. So I will just back out most of this and start over with MRP versions from the UC, I guess.

distribution/environments/aws-ec2-cloud/config/as-code/artifact-manager-s3.yaml
jglick: Well when that is merged, then I can see whether deliberately bogus plugin definitions and/or YAML configuration indeed result in failures.

jglick:
> Did `make container-check` run successfully for you locally?

Fails with something seemingly unrelated:

```
Creating network "services_default" with the default driver
Starting services_db_1 ... error

ERROR: for services_db_1  Cannot start service db: network 5ac4ecc334760a91bfa20dd325853d24e32e560aef07697719914214b9626093 not found

ERROR: for db  Cannot start service db: network 5ac4ecc334760a91bfa20dd325853d24e32e560aef07697719914214b9626093 not found
ERROR: Encountered errors while bringing up the project.
```

jglick:
Hmm, after `make -C distribution clean` then `make -C distribution container-check` passed the first time (after rerunning `npm install`, sigh). Then I ran it again, and it failed, but with an unrelated error:

```
test_jep_307
ASSERT:jenkins.io should be OK expected:<0> but was:<6>
…
FAILED (failures=1)
Makefile:32: recipe for target 'base-container-check' failed
make: *** [base-container-check] Error 1
```

So the tests seem a bit flaky still.

jglick:
Passes tests with #146 changes merged in.

batmat:
Possibly the local flakiness should be largely helped by https://github.com/jenkins-infra/evergreen/pull/145 if the issue you can see in your build/test-run... files relate to some plugin whose download failed.

batmat:
Retriggering, again. Last failure was for ci.j.i security maintenance.

PR #145, opened 2018-07-17
[JENKINS-51934] Retry plugin downloads when failing

batmat:
[JENKINS-51934](https://issues.jenkins-ci.org/browse/JENKINS-51934)

This PR replaces the existing `node-fetch` based code for downloading plugins, by a [`request-promise`](https://www.npmjs.com/package/request-promise) wrapper: `promise-request-retry`, which integrates the former and the _standard_ [`retry`](https://www.npmjs.com/package/retry) library.

### Automated testing 

* The first commit adds a test and checks the data is downloaded as expected (by checking checksum)

### Manual testing

I ran both `make unit`, and the much more global `make container-check`, while randomly disabling my (slow) WiFi connection. I saw the _retry count_ related log show up, then the download be retried successfully a few seconds after.

(Note: the `retry` default behavior is to use an exponential backoff strategy, starting with one second, then doubling each time, see https://www.npmjs.com/package/retry#retrytimeoutsoptions).

### DIFF
--- a/distribution/client/src/lib/downloader.js
+++ b/distribution/client/src/lib/downloader.js
@@ -7,7 +7,7 @@ const fs     = require('fs');
 const path   = require('path');
 const url    = require('url');
 
-const fetch   = require('node-fetch');
+const rp      = require('promise-request-retry');
 const logger  = require('winston');
 const mkdirp  = require('mkdirp');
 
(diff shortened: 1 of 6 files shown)


batmat:
Weird, failed right from the beginning as happens sometimes. Unrelated failure https://ci.jenkins.io/job/Infra/job/evergreen/job/PR-145/6/artifact/distribution/build/tests-run-2018-07-17_13-53-55-docker-cloud-tests.sh/distribution_instance_1.log

retriggering.

batmat:
```
info: Download complete /evergreen/jenkins/home/index.html

FAIL test/downloader.test.js (10.521s)

  ● Console



    console.log src/lib/__mocks__/fs.js:4

      Using a memfs filesystem...



  ● the Downloader class › download() › should return promise



    Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.

      

      at node_modules/jest-jasmine2/build/queue_runner.js:68:21

      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:628:19)
```

:confused: the download seems successful, but the async callback would not get called?! _Maybe_ this would be the case if the promise actually completes before reaching the `await response;` line?

distribution/client/src/lib/downloader.js
```diff
@@ -27,20 +27,38 @@ class Downloader {
...
 
-    return new Promise((resolve, reject) => {
-      fetch(item).then((res) => {
-        const output = fs.createWriteStream(filename);
+    let options = {
+      uri: item,
+      verboseLogging: true,
+      method: 'GET',
+      headers: {
+        'User-Agent': 'evergreen-client'
```
jglick: Should probably mention Jenkins somewhere, no?

batmat: Good question. I actually almost wrote something like "Jenkins-Essentials", and went back to evergreen-client, following somehow https://github.com/jenkinsci/jep/tree/master/jep/301#evergreen-client. 
I expect that, anyway, this should always be hitting Jenkins Project's related infra, so it should always be obvious what it's referring to, but well. Likely `jenkins-evergreen-client` would make it crystal clear indeed.

Will change it to that tomorrow, I think. No strong opinion.

Any opinion on this @rtyler.

rtyler: There's no need to make the user agent more explicit

distribution/client/src/lib/downloader.js
```diff
@@ -27,20 +27,38 @@ class Downloader {
...
+      },
+      simple: true,
+      resolveWithFullResponse: true,
+      encoding: null,
+      timeout: 10 * 1000,
+      retry: 3 // make it configurable?
```
rtyler: There's no need for end-user configuration of something like this, we're trying to get away from the infinite-configuration-switches remember :smiling_imp:

rtyler (approved):
Excellent work mon frer

PR #149, opened 2018-07-18
[JENKINS-52615] Set up an HTTP proxy cache for Essentials dev

batmat:
[JENKINS-52615](https://issues.jenkins-ci.org/browse/JENKINS-52615)

This change transparently sets up an `squid` proxy to use when the `evergreen-client` is downloading plugins. I have explicitly disabled the caching in CI, and have tested this locally.

The result is that with a preloaded volume, I was able to go from 0 to a started Jenkins Essentials instance in ~15 seconds.

### DIFF
--- a/distribution/Makefile
+++ b/distribution/Makefile
@@ -4,6 +4,7 @@ DOWNLOAD=curl -sSL
 # This variable is used for downloading some of the "upstream" maintained
 # scripts necessary for running Jenkins nicely inside of Docker
 SCRIPTS_URL=https://raw.githubusercontent.com/jenkinsci/docker/master/
+SQUID_VOLUME_NAME=squid-cache
 
 ### Phony targets
 #################
@@ -53,7 +54,7 @@ publish: container
 	$(MAKE) -C environments $@
 
 create-squid-cache-volume:
-	docker volume create squid-cache
+	docker volume create ${SQUID_VOLUME_NAME}
 
 run: create-squid-cache-volume
 	$(COMPOSE) up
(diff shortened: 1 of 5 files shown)


distribution/docker-compose.squid-cache.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
```
rtyler: My understanding is that this is a compose file override file which will be incorporated in `tests/utilities` for test set up?

If so, I recommend adding some helpfully explanatory comments to this file to make that relationship clear in a follow up set of changes.

distribution/Makefile
```diff
@@ -52,7 +52,10 @@ publish: container
...
 	$(MAKE) -C environments $@
 
-run:
+create-squid-cache-volume:
+	docker volume create squid-cache
+
+run: create-squid-cache-volume
 	$(COMPOSE) up
 
 clean:
```
rtyler: Please add some clean up for the volume to `make clean`

distribution/docker-compose.squid-cache.yml
batmat: I named it as clearly as I could, but right. Doing.

distribution/Makefile
batmat: done

distribution/docker-compose.squid-cache.yml
batmat: done

batmat:
@rtyler I believe your comments are addressed.

rtyler (approved):
Merge when Jenkins says it's ready :+1:

PR #152, opened 2018-07-18
Configure the sequelize CLI based off the existing JSON configuration…

rtyler:
… files

This introduces a new supported environment variable DB_CONNECTION_STRING, which
should be used in production environments (e.g. Kubernetes) for specifying a
custom authenticated database endpoint


batmat:
Sigh. I think I introduced flakiness in Jest `downloader.test.js`. I _think_ I've seen this error once or twice, random.

PR #153, opened 2018-07-19
Use pipeline-github-notify-step plugin to make feedback quicker

batmat:
Analyzing the failure cause can be a bit time-consuming. I have used https://plugins.jenkins.io/pipeline-githubnotify-step successfully in the past for enriching the PR statuses, and understand quicker what went wrong.

I am totally aware this adds quite some verbosity in the `Jenkinsfile`, but I think it's OK and is going to be pretty helpful.

**WIP**

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -108,16 +108,16 @@ pipeline {
                 stage('Base image') {
                   agent { label 'linux' }
                   steps {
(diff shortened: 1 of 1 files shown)


batmat:
@rtyler I do not fully disagree :-).

I think it's valuable to quickly see what flavor failed to be tested successfully. Now the `Verifications` and `Build Images` are probably better not detailed, as anyway they do not fail that often in CI. Going to change this.

PR #154, opened 2018-07-19
Create a container for running migrations on deployment

rtyler:
See [JENKINS-52526](https://issues.jenkins-ci.org/browse/JENKINS-52526)


rtyler:
@batmat merge it before Jenkins changes its mind! :smiling_imp:

batmat:
@rtyler looks good to me, but I wonder if we couldn't actually test this container, by:
* adding `container-migration` in dependency of the `migrate` target
* use the locally built image `jenkinsciinfra/evergreen-backend-migration` and run it in replacemenf of the current code.

Possibly, this would create a chicken-and-egg build issue though, didn't check, yet.

PR #165, opened 2018-08-01
Add a `status ping` event to help maintain our socket.io connection

rtyler:


services/src/app.js
```diff
@@ -128,4 +128,9 @@ app.configure(jwt({
   }
 }));
 
+setInterval(() => {
```
batmat: Couldn't we use the `cron` service, instead of manually using `setInterval`?

services/src/app.js
```diff
@@ -128,4 +128,9 @@ app.configure(jwt({
...
+  logger.info('Sending ping');
+  app.service('status').emit('ping', { 'timestamp' : (new Date()) });
```
batmat: AIUI `Date.now()` seems now preferrable to `new Date()`, since ES 5.1. As it does not need creating a new object.

services/src/app.js
rtyler: This is in the services which don't have the cron package built in, I'm not going to add that dependncy when setInterval is built in already for just this one function :smile:

PR #167, opened 2018-08-02
[Jenkins 51766] instance flavor

mandie722:
Flavor is now a required field when calling status.create.

### DIFF
--- a/distribution/client/src/client.js
+++ b/distribution/client/src/client.js
@@ -24,6 +24,10 @@ const Update         = require('./lib/update');
  */
 class Client {
   constructor() {
+    if (!process.env.FLAVOR) {
+      logger.error('Fatal error encountered while trying to start, no flavor set, exiting the client');
+      throw new Error('Missing flavor definition');
+    }
     this.app = feathers();
     this.reg = new Registration(this.app);
     this.status = new Status(this.app, {flavor: process.env.FLAVOR});
--- a/distribution/client/src/client.js
+++ b/distribution/client/src/client.js
@@ -26,7 +26,7 @@ class Client {
   constructor() {
     this.app = feathers();
     this.reg = new Registration(this.app);
-    this.status = new Status(this.app);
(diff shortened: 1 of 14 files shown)


distribution/client/src/client.js
```diff
@@ -26,7 +26,7 @@ class Client {
   constructor() {
     this.app = feathers();
     this.reg = new Registration(this.app);
-    this.status = new Status(this.app);
+    this.status = new Status(this.app, {flavor: process.env.FLAVOR});
```
batmat: I think we must have an explicit check of `process.env.FLAVOR` above, so that we can bail out fast, with a crystal clear message. It's IMO going to be useful in the very near term for us as we battle-test this awesome improvement, but immediately stopping the client for users running into crazy issues, or trying wild customization would help everyone avoid losing a lot of time figuring this env var is actually missing for whatever reason.

mandie722: how do we bail in a useful way here?  just throw an exception?

rtyler: That seems reasonable to me. An uncaught exception will likely cause the client not to start properly

batmat: BTW, reading https://ci.jenkins.io/job/Infra/job/evergreen/job/PR-167/4/artifact/distribution/build/tests-run-2018-08-03_17-19-57-no-flavor-tests.sh/distribution_instance_1.log I now think there should also be an additional unconditional log about the flavor in use during startup.

For instance, again about https://ci.jenkins.io/job/Infra/job/evergreen/job/PR-167/4/artifact/distribution/build/tests-run-2018-08-03_17-19-57-no-flavor-tests.sh/distribution_instance_1.log I do not even understand how it is able to start up: the `base-container-check` should now always fail, since by definition this one has no flavor.

Can you clarify how/why this works @mandie722 ? Thanks

mandie722: I don't think it's working as expected.  From what I'm seeing, no flavor is actually using the docker-cloud configuration.

PR #168, opened 2018-08-03
[JENKINS-52728] Hide FreeStyle and Native Maven jobs 

batmat:
[JENKINS-52728](https://issues.jenkins-ci.org/browse/JENKINS-52728)

The real gist of this change (to hide some types of Job from the `/newJob` Jenkins page) is brought by changes in the _Essentials Jenkins plugin_ in https://github.com/jenkinsci/essentials-plugin/pull/10 and https://github.com/jenkinsci/essentials-plugin/pull/11.

---

h3. Why apparently unrelated changes disabling the setup wizard?

You can see that this PR also includes a change that:

* disables the upgrade wizard
* adds an admin user with a random password

This has been added to be able to automatically test these TopLevelItems were *actually* filtered using the existing `tests/tests.sh` shUnit2 based tests.

But because of the wizard, attempts to hit Jenkins endpoints during automated tests were being redirect to the setup wizard. So I [asked the dev ML for feedback](https://groups.google.com/d/msg/jenkinsci-dev/toKqQkaMHBw/rczG4--QCgAJ) about the expected UX, and also added this in the go to be then able to write these automated tests and wrap up the whole story.

h3. Why the hell did I add an XML API on the Essentials plugin in https://github.com/jenkinsci/essentials-plugin/pull/11 ?

Because trying to write these tests, I thought initially I would just call `/newJob` and parse the HTML there, but I discovered working on this that this page is actually Javascript generated, hence I couldn't use it using a simple `curl` call to check what were the different item types I was actually allowed to create... So I wrote an XML endpoint in https://github.com/jenkinsci/essentials-plugin/pull/11 so that I could simply parse that XML using shell script and check the native Maven, and FreeStyle jobs descriptors weren't present...

Hope this clarifies.

---


batmat:
WUT

image
image

distribution/config/as-code/create-admin-user.yaml
```diff
@@ -0,0 +1,10 @@
+jenkins:
+  remotingSecurity:
+    enabled: false
+  securityRealm:
+    local:
+      allowsSignup: false
+      users:
+        - id: "admin"
+          password: ${ADMIN_PASSWORD:-essentials}
```
rtyler: Rather than setting all instances to the same password, would it be possible for a startup script to generate a password for the instance and then set this environment variable?

distribution/Dockerfile
```diff
@@ -23,7 +23,8 @@ ENV JAVA_OPTS=\
 "-Djenkins.model.Jenkins.workspacesDir=${JENKINS_VAR}/jobs/\${ITEM_FULL_NAME}/workspace "\
 "-Djenkins.model.Jenkins.buildsDir=${JENKINS_VAR}/jobs/\${ITEM_FULL_NAME}/builds "\
 "-Dhudson.triggers.SafeTimerTask.logsTargetDir=$JENKINS_VAR/logs "\
-"-Djava.util.logging.config.file=$EVERGREEN_HOME/logging.properties "
+"-Djava.util.logging.config.file=$EVERGREEN_HOME/logging.properties "\
+"-Djenkins.install.runSetupWizard=false "
```
rtyler: :clap:

distribution/config/as-code/create-admin-user.yaml
batmat: This is the idea. I agree it's not done yet, but I was definitely planning to not use that defaulting for the final target. It was just easier here for the test setup. And also always having the same password locally for testing seems nice (but can easily be achieved by setting up docker compose correctly).

I agree it must be done rather sooner than later though. Will have a look

rtyler: I don't want to merge this change until that's done, because we're perilously close to having a version ready for alpha testers.

I would rather alpha testers have secure instances.

jglick:
The PR title seems rather misleading.

distribution/scripts/jenkins-essentials.sh
```diff
@@ -0,0 +1,25 @@
...
+passwordFileLocation="$JENKINS_HOME/secrets/initialAdminPassword"
+passwordFileLocationDirectory="$( dirname "$passwordFileLocation" )"
+
+echo -n "Creating $passwordFileLocationDirectory... "
+mkdir -p "$passwordFileLocationDirectory"
+echo "Done."
+
+echo -n "Generating admin password... "
+echo $RANDOM | md5sum | cut -d ' ' -f 1 > "$passwordFileLocation"
+echo "Done. Password value stored in $passwordFileLocation file."
```
jglick: How does the user actually get access to that file? Do they need to `docker-cp` or something?

distribution/scripts/jenkins-essentials.sh
```diff
@@ -0,0 +1,25 @@
...
+
+JENKINS_ADMIN_PASSWORD="$( cat "$passwordFileLocation" )"
+
+# Intended for ease of development. By default, password is obviously *not* put in logs.
+if [[ -n $INSECURE_SHOW_ADMIN_PASSWORD ]]; then
```
jglick: Not `= yes` ? Double-quotes around that?

distribution/scripts/jenkins-essentials.sh
```diff
@@ -0,0 +1,25 @@
...
+  echo "[WARNING] INSECURE_SHOW_ADMIN_PASSWORD defined, it should only ever be done for testing."
```
jglick: But I thought you were now doing this by default, acc. to `docker-compose.yml`?

services/ingest-update-center
```diff
@@ -32,7 +32,7 @@ let ingest = {
      {
        groupId: 'io.jenkins.plugins',
        artifactId: 'essentials',
-       url: 'http://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/essentials/0.3-rc20.a3f76cdd1348/essentials-0.3-rc20.a3f76cdd1348.hpi',
+       url: 'https://repo.jenkins-ci.org/incrementals/io/jenkins/plugins/essentials/0.3-rc22.db3e99966ef0/essentials-0.3-rc22.db3e99966ef0.hpi',
```
jglick: Nice to link to the corresponding upstream PR.

distribution/scripts/jenkins-essentials.sh
batmat: Docker-compose is exclusively used for dev, not for production deployment (where one would only run the instance, not the backend services)

distribution/scripts/jenkins-essentials.sh
batmat: I do not *really* care about the value, only checking if that env car is defined actually.

distribution/scripts/jenkins-essentials.sh
batmat: As it is, `docker exec cat …` seems slightly simpler, but basically yes, that is the idea.

batmat:
Also, @jglick, about:
> The PR title seems rather misleading.

This was, and still really is, the intent of the change here. I suspect you're referring to the fact I'm also creating an admin user and disabling the setup wizard. I just updated the description to explain the reason: https://github.com/jenkins-infra/evergreen/pull/168#issue-206012607 Hope this helps understand it better.

batmat:
@rtyler your request about creating random passwords for each instance has been addressed, care to have another look? Thanks

distribution/scripts/jenkins-essentials.sh
jglick: Just a bit of technical debt for the next person to try to pass `=no` and be confused when this is treated like `=yes`.

batmat:
@rtyler ping. Thanks

PR #169, opened 2018-08-03
Rely on essentials.yaml and hpi-specified dependencies for generating ingest.json, etc

rtyler:


rtyler:
@batmat Perhaps you're better at parsing the logs than I am, I was only seeing a Docker daemon connection failure. Did you spot a more accurate error?

mandie722:
It took some fine tooth comb parsing, but this looks like the failure:

test_required_plugins_are_here

ls: /evergreen/jenkins/home/plugins/metrics.hpi: No such file or directory

ASSERT:The metrics plugin should be installed expected:<0> but was:<1>

rtyler:
@mandie722 thanks for the catch,  interesting :smile:

services/essentials.yaml
```diff
@@ -6,18 +6,19 @@ metadata:
...
-      version: '0.1-alpha'
+      version: '0.11-alpha-rc362.942711740b07'
+    - groupId: 'io.jenkins.plugins'
+      artifactId: 'essentials'
+      version: '0.3-rc20.a3f76cdd1348'
 
     - groupId: 'io.jenkins.blueocean'
       artifactId: 'blueocean'
-      version: '1.4.2'
+      version: '1.7.1'
```
jglick: Last I saw, this pulled latest dependencies from the update center, and so was quite useless. Still true?

services/ingest.json
```diff
@@ -1,894 +1,930 @@
...
-        "signature": "e1cdf271eabed428c2198c4f7a49c5dea7fc8850987bcf63b87b74936a3cc426896a4b7c9c974f0fe3abd61071ea5a2238538af9f9266f2d37696856809d697a"
+        "type": "sha256",
+        "signature": "392f0a30de31dadcaa2a06b8e07ff721bcf8fe37eb520d5a669dd5a8cdb17301"
       }
     },
     {
       "groupId": "io.jenkins.blueocean",
       "artifactId": "blueocean",
-      "url": "http://updates.jenkins-ci.org/download/plugins/blueocean/1.7.1/blueocean.hpi",
+      "url": "https://repo.jenkins-ci.org/releases/io/jenkins/blueocean/blueocean/1.7.1/blueocean-1.7.1.hpi",
```
jglick: This looks like a mistake.

services/ingest.json
```diff
@@ -1,894 +1,930 @@
...
-      "url": "http://updates.jenkins-ci.org/download/plugins/workflow-scm-step/2.6/workflow-scm-step.hpi",
+      "url": "https://updates.jenkins.io/download/plugins/workflow-scm-step/2.6/workflow-scm-step.hpi",
       "checksum": {
         "type": "sha256",
         "signature": "o4KGStrw/IDVgkNWgVehRQAo5tBQOgO4ERHUxN6CddA="
       }
     },
     {
-      "groupId": "org.jenkins-ci.plugins.workflow",
+      "groupId": "unknown",
```
jglick: Huh?

services/ingest.json
```diff
@@ -1,894 +1,930 @@
...
     {
-      "groupId": "org.jenkins-ci.plugins",
+      "groupId": "io.jenkins",
       "artifactId": "configuration-as-code",
       "url": "https://repo.jenkins-ci.org/incrementals/io/jenkins/configuration-as-code/0.11-alpha-rc362.942711740b07/configuration-as-code-0.11-alpha-rc362.942711740b07.hpi",
       "checksum": {
-        "type": "sha512",
-        "signature": "4faefcc46299557efaa1c896835a2413ad0646201e40938b4207dbd4cfecf33303cf5f28fd1f53cb71146c45d12c7fa34d2d87d312acf9fd7964c08506d2278d"
+        "type": "sha256",
+        "signature": "5e0d07f6bc86ea23dbe6f2b40d15fe8759aa43470a45d9ee09d90f415e007575"
```
jglick: Why is this hex and others are Base-64?

services/ingest.json
```diff
@@ -897,29 +933,23 @@
...
-          "groupId": "org.jenkins-ci.plugins",
+          "groupId": "unknown",
           "artifactId": "ssh-slaves",
-          "url": "http://updates.jenkins-ci.org/download/plugins/ssh-slaves/1.26/ssh-slaves.hpi",
-          "checksum": {
-            "type": "sha256",
-            "signature": "az0y+gKCp0nc8q09rG54B1hTPgK3MbqHhVo5cs9ncC0="
-          }
+          "url": "https://updates.jenkins.io/download/plugins/ssh-slaves/1.26/ssh-slaves.hpi",
+          "checksum": {}
```
jglick: No checksum?

services/process-dependencies
```diff
@@ -0,0 +1,227 @@
...
+// Perform a first pass on the plugins needed for all Evergreen installations
+const firstPass = essentials.spec.plugins.map(plugin => processPlugin(plugin, ingest.plugins));
+
+// Collection of plugins which are needed but not specified in essentials.yaml
+let needed = {};
+
+
+const resolvePlugin = (name, inventory) => {
+  // We don't have the plugin in essentials.yaml, so it's dependencies
+  // should be resolved through the update center unless otherwise
```
jglick: This is a bad policy—nondeterministic! Use the least upper bound instead.

services/src/services/update/update.class.js
```diff
@@ -119,6 +119,7 @@ class Update extends FeathersSequelize.Service {
     record.manifest.plugins.forEach((plugin) => {
       computedManifest.plugins.updates.push({
         url: plugin.url,
+        artifactId: plugin.artifactId,
```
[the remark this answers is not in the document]
rtyler: __

services/process-dependencies
```diff
@@ -0,0 +1,227 @@
...
+    logger.info(`Fetching ${pluginUrl} took ${Date.now() - start}`);
+    return processManifest(response);
+  } catch (err) {
+    if (err.statusCode == 404) {
+      logger.error(`Could not find ${pluginUrl}`);
+    }
+    else {
+      logger.warn(err);
+    }
+    return null;
```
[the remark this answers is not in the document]
rtyler: Agreed, it's worth dying on these errors.

services/essentials.yaml
jglick: Well, when I worked on my last PR (#144), I noticed that the transitive dependencies were taken from the latest version on the UC. Since the `blueocean` plugin itself is just an aggregator, that made the actual version you specify here pretty much irrelevant—running the ingestion script would always pick up the latest version of every component. I have not tried the rewritten script locally, but I had another comment to the effect that you should never use a nondeterministic version for a transitive dependency: you should use least upper bounds.

services/process-dependencies
rtyler: Please help me understand why pulling the latest released version of plugins is not a good thing for the mission of Jenkins Essentials here.

@batmat the desire for testing you describe is still easily done with providing a mock update center during test.

We can chat more about the resolution behavior in the meeting this morning

services/ingest.json
jglick: Sorry, was on mobile.

You are switching from a (presumably scalable) Jenkins-run official update site to an Artifactory repository, for a release version of a plugin. IMO there should be no `url`s using https://repo.jenkins-ci.org/releases/ since these could all be served by http://updates.jenkins-ci.org/download/plugins/ instead.

services/process-dependencies
jglick: > not dependent on new releases in UC

In particular I see no reason why the script should load the UC JSON at all. It should be irrelevant.

Ideally it would use the Artifactory API to parse manifests except in cases where the plugin already exists in the local filesystem in which case it can more efficiently load that.

services/process-dependencies
```diff
@@ -0,0 +1,227 @@
...
+      });
+    });
+  return inventory;
+};
+
+Promise.all(firstPass).then(response => filterPluginsDependencies(response, needed))
+.then((inventory) => {
+  Object.keys(inventory).forEach((name) => {
+    ingest.plugins.push({
+        groupId: 'unknown',
```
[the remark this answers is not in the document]
jglick: Right…if the `groupId` is used by the client, then we ought to get it right. If it is not used by the client (and I can think of no reason why it would be…?) then the field should be deleted from the output JSON.

services/src/services/update/update.class.js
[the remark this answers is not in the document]
jglick: The official term is `shortName` as [seen here](https://github.com/jenkinsci/jenkins/blob/7b2698db89183b4d701107485db61eb36442ac25/core/src/main/java/hudson/PluginWrapper.java#L147-L153) and [here](https://github.com/jenkinsci/jenkins/blob/7b2698db89183b4d701107485db61eb36442ac25/core/src/main/java/hudson/PluginWrapper.java#L343-L346). That is what Jenkins itself groks. `artifactId` is a Maven concept, and it is `maven-hpi-plugin` which causes a generated `*.hpi` to be given a short name matching the artifact ID by default.

From the perspective of generated JSON, i.e. the client, you should not care about Maven.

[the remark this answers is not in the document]
jglick: FTR in `update-center.actual.json`, the plugin short name is apparently both the key in a map and a `name` attribute, e.g.:

```json
    "workflow-api": {
      "name": "workflow-api",
```

services/process-dependencies
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env node
+
+const fs      = require('fs');
+const path    = require('path');
+const yaml    = require('js-yaml');
+const logger  = require('winston');
+const request = require('request-promise');
+const cheerio = require('cheerio')
```
rtyler: This is no longer being used and should be removed :fire:

services/process-dependencies
rtyler: Some options described by @jglick 

1. Make `essentials.yaml` contain every single plugin in the graph
1. Load in transitive dependencies by collecting the listing of depX and select the highest version specified.

Another option: use route 1 with some automation to check for updates, e.g. `workflow-job`, and submit a pull request to `essentials.yaml`

rtyler:
So I've got some work in progress that completely reworks this, I'm moving pretty swiftly along with it, but I don't think I will be able to refactor it in time for @batmat in the morning.

I'll try to get the refactored resolver tool ready before the end of wednesday for mon frer :smile_cat:

PR #176, opened 2018-08-14
Ensure that out-of-sync essentials.yaml specs cannot be committed

rtyler:
This means that spec and status must always be in sync in the tree. :+1:

Fixes JENKINS-53041


rtyler:
Going to go ahead and merge this to prevent bad commits, IMHO it's fairly iinnocuous.

PR #190, opened 2018-08-20
Lower log level

batmat:
Using https://github.com/jenkinsci/essentials-plugin/pull/12 (to avoid sending this to Sentry)

cc @jglick


rtyler:
Apparently Jenkins just never fired here. WTF.

Re-opening to trigger this bad boy

rtyler:
Hooray, Jenkins worked correctly!

PR #198, opened 2018-08-24
Add an acceptance test for socket.io events published from update.create()

rtyler:


rtyler:
Huh, somehow the issue I was seeing isn't just local. For whatever reason changing this has resulted in the FLAVOR apparently not being set properly when running the docker-cloud test

PR #202, opened 2018-08-25
[JENKINS-53147] Rebrand to Evergreen: use rebranded Jenkins plugin

batmat:
[JENKINS-53147](https://issues.jenkins-ci.org/browse/JENKINS-53147)

Using https://github.com/jenkinsci/evergreen-plugin/pull/13

### DIFF
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -11,7 +11,10 @@ spec:
       version: 0.11-alpha-rc362.942711740b07
     - groupId: io.jenkins.plugins
       artifactId: evergreen
-      version: 0.3-rc24.69ee85fa5e9a
+      version: 1.0-rc50.78d526a2ba0f
+    - groupId: io.jenkins.plugins
+      artifactId: essentials
+      version: 0.5.1
     - groupId: io.jenkins.blueocean
       artifactId: blueocean
       version: 1.8.2
@@ -217,8 +220,11 @@ status:
       artifactId: ec2
       version: '1.39'
     - groupId: io.jenkins.plugins
+      artifactId: essentials
+      version: 0.5.1
+    - groupId: io.jenkins.plugins
       artifactId: evergreen
-      version: 0.3-rc24.69ee85fa5e9a
+      version: 1.0-rc50.78d526a2ba0f
     - groupId: org.jvnet.hudson.plugins
       artifactId: favorite
       version: 2.3.1
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -10,8 +10,8 @@ spec:
       artifactId: configuration-as-code
       version: 0.11-alpha-rc362.942711740b07
     - groupId: io.jenkins.plugins
-      artifactId: essentials
-      version: 0.3-rc23.7230a0298a47
+      artifactId: evergreen
+      version: 0.3-rc24.69ee85fa5e9a
     - groupId: io.jenkins.blueocean
       artifactId: blueocean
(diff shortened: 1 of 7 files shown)


services/essentials.yaml
```diff
@@ -10,8 +10,8 @@ spec:
       artifactId: configuration-as-code
       version: 0.11-alpha-rc362.942711740b07
     - groupId: io.jenkins.plugins
-      artifactId: essentials
-      version: 0.3-rc23.7230a0298a47
+      artifactId: evergreen
```
rtyler: Is it legal to have both of these plugins installed at once? I'm imagining what would happen with an existing instance when this update level is pushed out.

batmat: Good point I overlooked. I _seem_ to remember some discussion around this, but cannot find it now. Also, nothing expressed about that in [BoM's JEP 309](https://github.com/jenkinsci/jep/tree/master/jep/309), I'll have a deeper look tomorrow I think.

We probably should have some way to actually say that a given plugin is *not* here for upgradeability purposes in the future IMO.

For now, I think I'm going to handle this by releasing an empty (aka _tombstoned_) essentials-plugin.

cc @oleg-nenashev for the BoM aspect, FYI.

oleg-nenashev: JEP-309 does not define behavior for Evergreen specifically. It's up to the BOM packaging logic to define and enforce behavior it needs, BOM just provides the documentation format. 

Tombstoning plugins and declaring dependencies on the new ones is what we actually do when there is an unavoidable need to change artifactId.

rtyler: @batmat the Update Manifest format has space for `upgrades` and `deletes`, we could look into using this as an opportunity to introduce the deletes :smile:

batmat: Makes sense to me. See https://issues.jenkins-ci.org/browse/JENKINS-53164?focusedCommentId=347263&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-347263

batmat: Tombstone release just done, through https://github.com/jenkinsci/evergreen-plugin/pull/14 and then released from my machine:
https://repo.jenkins-ci.org/releases/io/jenkins/plugins/essentials/0.4-essentials-tombstoned/

services/essentials.yaml
```diff
@@ -9,9 +9,12 @@ spec:
     - groupId: io.jenkins
       artifactId: configuration-as-code
       version: 0.11-alpha-rc362.942711740b07
+    - groupId: io.jenkins.plugins
+      artifactId: evergreen
+      version: 1.0-rc50.78d526a2ba0f
     - groupId: io.jenkins.plugins
       artifactId: essentials
-      version: 0.3-rc23.7230a0298a47
+      version: 0.5.1
```
batmat: Tombstoned release for `essentials` artifactId, i.e. empty plugin.

batmat:
So, concern about `essentials` and `evergreen` plugin just addressed: I've released a "tombstoned" version of the `essentials` plugin. I.e. one where I removed basically everything: the `.hpi` contains almost nothing. Hence, it can be installed next to `evergreen` Jenkins plugin without issue.

If one wants to check the `essentials` is actually empty: http://repo.jenkins-ci.org/releases/io/jenkins/plugins/essentials/0.5.1/essentials-0.5.1.hpi

batmat:
Seems like a pure `npm` download issue unrelated to the change, retriggering.

PR #203, opened 2018-08-27
Providing feedback on the booting process to users

rtyler:
There's still some more work I need to do here to build a little status app inside `evergreen-client`


distribution/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:9-alpine
+FROM node:10-alpine
```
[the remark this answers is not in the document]
rtyler: Exactly, I've thought about this kind of thing a lot over the past couple weeks. Part of the motivation to get this user feedback system is to allow us to inform the users a new container being present (hopefully)

README.adoc
```diff
@@ -45,7 +45,7 @@ docker volume create jenkins-evergreen-data && \
 docker run --name evergreen \
     --restart=always \
     -ti \
-    -p 8080:8080 \
+    -p 8080:80 \
```
batmat: Interesting issue I just hit because of this: this docs got updated immediately, when for some reason, the public Docker image seems to be still not updated for some reason. So Running the docs as is was binding 8080 to 80, but my local image does not have `nginx` yet. (yes, I did a `docker pull` in case)

batmat: FTR, the updated image has been updated now.

PR #205, opened 2018-08-28
 [JENKINS-53285] Update plugins to latest to avoid warnings

batmat:
https://issues.jenkins-ci.org/browse/JENKINS-53285 this should already reduce the number of warnings we receive in Sentry.

cc @jglick

### DIFF
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -90,12 +90,6 @@ spec:
     - groupId: org.jenkins-ci.main
       artifactId: maven-plugin
       version: 3.1.2
-    - groupId: org.jenkins-ci.plugins
-      artifactId: node-iterator-api
-      version: 1.5.0
-    - groupId: org.jenkins-ci.plugins
-      artifactId: run-condition
-      version: '1.0'
   environments:
     - name: docker-cloud
       plugins:
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -69,6 +69,38 @@ spec:
     - groupId: io.jenkins.plugins
       artifactId: artifact-manager-s3
       version: '1.1'
+      # transitive dependencies that we want to force the upgrade of
+      # Note: it would be nice to have a way to fail the evergreen build if
+      # those things below are *not* already pulled in transitively and just
+      # the version is bumped
+    - groupId: org.jenkins-ci.plugins
+      artifactId: authentication-tokens
+      version: '1.3'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: conditional-buildstep
+      version: '1.3.6'
+    - groupId: org.jenkins-ci.plugins.icon-shim
+      artifactId: icon-shim
+      version: '2.0.3'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: javadoc
+      version: '1.4'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: jsch
+      version: '0.1.54.2'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: mapdb-api
+      version: '1.0.9.0'
+    - groupId: org.jenkins-ci.main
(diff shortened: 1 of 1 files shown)


services/essentials.yaml
```diff
@@ -69,6 +69,33 @@ spec:
     - groupId: io.jenkins.plugins
       artifactId: artifact-manager-s3
       version: '1.1'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: authentication-tokens
+      version: '1.3'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: conditional-buildstep
```
rtyler: This is no longer required and shouldn't be here.

services/essentials.yaml
```diff
@@ -69,6 +69,33 @@ spec:
...
+      version: 1.3.6
+    - groupId: org.jenkins-ci.plugins.icon-shim
+      artifactId: icon-shim
+      version: 2.0.3
+    - groupId: org.jenkins-ci.plugins
+      artifactId: javadoc
```
rtyler: This is also no longer required.

services/essentials.yaml
```diff
@@ -69,6 +69,33 @@ spec:
...
+      version: '1.4'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: jsch
+      version: 0.1.54.2
+    - groupId: org.jenkins-ci.plugins
+      artifactId: mapdb-api
```
rtyler: This is also no longer required

services/essentials.yaml
```diff
@@ -69,6 +69,33 @@ spec:
...
+      version: 1.0.9.0
+    - groupId: org.jenkins-ci.main
+      artifactId: maven-plugin
+      version: 3.1.2
+    - groupId: org.jenkins-ci.plugins
+      artifactId: node-iterator-api
+      version: 1.5.0
+    - groupId: org.jenkins-ci.plugins
+      artifactId: run-condition
```
rtyler: This is also no longer required

services/essentials.yaml
rtyler: Looks like this is required by the newer maven plugin

services/essentials.yaml
```diff
@@ -69,6 +69,21 @@ spec:
...
+      artifactId: icon-shim
+      version: 2.0.3
+    - groupId: org.jenkins-ci.plugins
+      artifactId: javadoc
+      version: '1.4'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: jsch
+      version: 0.1.54.2
+    - groupId: org.jenkins-ci.main
+      artifactId: maven-plugin
```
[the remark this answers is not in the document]
batmat: I think I tried in the past, but not recently. The whole world was still depending on it, and I had filed https://issues.jenkins-ci.org/browse/JENKINS-46684 back then. I will need to recheck, possibly now the situation improved. Looking quickly.

services/essentials.yaml
```diff
@@ -69,6 +69,21 @@ spec:
     - groupId: io.jenkins.plugins
       artifactId: artifact-manager-s3
       version: '1.1'
+    - groupId: org.jenkins-ci.plugins
+      artifactId: authentication-tokens
+      version: '1.3'
+    - groupId: org.jenkins-ci.plugins.icon-shim
+      artifactId: icon-shim
```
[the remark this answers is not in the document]
batmat: So, I just tried, and it's pulled in by something else. (didn't check what yet). And if so, it's the 1.0.3 that get taken, so this must stay at least for now.

services/essentials.yaml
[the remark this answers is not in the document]
batmat: Yay! Enriched and repurposed a bit https://github.com/jenkins-infra/evergreen/pull/208 to include this. Thanks for checking Jesse!

PR #208, opened 2018-08-29
[JENKINS-53285] bump to fixed version of node-iterator-api plugin

batmat:
(This plugin is pulled in by the `ec2` plugin). 

[JENKINS-53285](https://issues.jenkins-ci.org/browse/JENKINS-53285)

Between `1.5` and `1.5.0`, the parent pom got updated. So we should not have this warning anymore using  `1.5.0`: https://github.com/jenkinsci/node-iterator-api-plugin/compare/7b284cb95b1ebcd09a430b94fd75d5599e867668...f73034788d1c8752ab7bef3af6aea664f3af674e

### DIFF
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -104,6 +104,9 @@ spec:
         - groupId: io.jenkins.plugins
           artifactId: artifact-manager-s3
           version: '1.1'
+        - groupId: org.jenkins-ci.plugins
+          artifactId: node-iterator-api
+          version: 1.5.0
 status:
   core:
     version: '2.138'


services/essentials.yaml
```diff
@@ -104,6 +104,9 @@ spec:
         - groupId: io.jenkins.plugins
           artifactId: artifact-manager-s3
           version: '1.1'
+        - groupId: org.jenkins-ci.plugins
+          artifactId: node-iterator-api
+          version: 1.5.0
```
batmat: pulled in by `ec2` plugin. We would ideally IMO need some way to tell the system we're only declaring a given plugin to override the version being pulled in transitively. IOW, in that case, this plugin should be uninstalled if it is not pulled in transitively anymore.

PR #213, opened 2018-08-30
Tyler integrity checks 53227

batmat:
#209 with `lint` failure fixed, to see if this goes to completion green with this fixed. Or we go to another more important error.


rtyler:
Cherry picked onto #209

PR #214, opened 2018-08-30
Fix make run

batmat:
Mainly, `make run` was failing because the `FLAVOR` is undefined.
So we now use the `docker-cloud` flavor for local testing.

I went on fixing that while starting to work on [JENKINS-53215](https://issues.jenkins-ci.org/browse/JENKINS-53215).

### DIFF
--- a/distribution/Makefile
+++ b/distribution/Makefile
@@ -59,8 +59,12 @@ publish: container
 create-squid-cache-volume:
 	docker volume create ${SQUID_VOLUME_NAME}
 
-run: create-squid-cache-volume
-	$(COMPOSE) -f docker-compose.yml -f docker-compose.squid-cache.yml up
+run: create-squid-cache-volume container-check-prereqs
+	$(COMPOSE) -f docker-compose.yml \
+             -f docker-compose.squid-cache.yml \
+             -f tests/environments/docker-cloud/docker-compose.docker-cloud.yml \
+             up
+
 
 clean:
 	$(COMPOSE) down || true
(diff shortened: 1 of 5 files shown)


rtyler:
The pull request description doesn't match up to the code I'm seeing. Did something not get committed @batmat ?

distribution/Makefile
```diff
@@ -59,8 +59,12 @@ publish: container
 create-squid-cache-volume:
 	docker volume create ${SQUID_VOLUME_NAME}
 
-run: create-squid-cache-volume
-	$(COMPOSE) -f docker-compose.yml -f docker-compose.squid-cache.yml up
+run: create-squid-cache-volume container-check-prereqs
+	$(COMPOSE) -f docker-compose.yml \
+             -f docker-compose.squid-cache.yml \
+             -f tests/environments/docker-cloud/docker-compose.docker-cloud.yml \
```
batmat: The gist of this is here @rtyler, to use an image that has a `FLAVOR` env var.

distribution/tests/utilities
```diff
@@ -27,7 +27,7 @@ cleanup () {
 
   echo -n "Retrieving test run data (container logs, ...) before shutdown... Will be in the $testRunDirectory directory... "
   # Normally should be something like evergreen_backend_1, evergreen_db_1 and evergreen_instance_1
-  containers=$( $COMPOSE ps | head -5 | tail -3 | awk '{print $1}' )
+  containers=$( $COMPOSE ps | head -5 | tail -4 | awk '{print $1}' )
```
batmat: *That* is indeed unrelated @rtyler, that might be the thing that was losing you?

PR #220, opened 2018-08-31
[JENKINS-53215] Configure Secure defaults OOTB

batmat:
[JENKINS-53215](https://issues.jenkins-ci.org/browse/JENKINS-53215)

Jenkins Evergreen should have no Admin monitor related to secure defaults showing under /manage after startup.

----
Note: for now, this does not fix the `Jenkins root URL is empty but is required for the proper operation of many Jenkins features like email notifications, PR status update, and environment variables such as BUILD_URL.` monitor:

image

This one has been fixed only very recently in the core itself, and requires us to guide users to the /configure page and save it, since we cannot know in advance what the value is going to be, at least for the Docker Cloud flavor.
We might be able to do it though for the AWS flavor.
That is not a security related issue however, so it's not very critical at this point.

### DIFF
--- a/distribution/config/as-code/create-admin-user.yaml
+++ b/distribution/config/as-code/create-admin-user.yaml
@@ -1,6 +1,7 @@
 jenkins:
+  # Agent to master security
   remotingSecurity:
-    enabled: false
+    enabled: true
   securityRealm:
     local:
       allowsSignup: false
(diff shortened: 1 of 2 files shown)


batmat:
I pushed a commit with a test, expected to fail. The second should fix the test. I will likely squash the two once I get the expected result (i.e. failure, then success).

batmat:
Going to merge this as I would like to test it too in real end-user conditions by pulling the image to be published. Not a very big change anyway and I've checked it locally a few times.

I will still welcome a look from you @mandie722 and @rtyler to see if I didn't miss something. If so, I'll address in a followup.

PR #226, opened 2018-09-05
Increase the timeout for some of our more flakey tests and restructure

rtyler:
It's still unclear why these are failing in trusted.ci, and _only_ in
trusted.co.

See:
    https://gist.github.com/rtyler/fa8dad5b4e23ac49fdf5ab6170439c75


batmat:
Given the public images are still 7 days old on Docker Hub, I suspect that unfortunately this didn't work.

PR #227, opened 2018-09-05
Ensure Jenkins doesn't redirect to http://localhost from localhost:8080

rtyler:
Unfortunately I could not find a good way to test this :-/

Fixes JENKINS-53345

### DIFF
--- a/distribution/config/nginx.conf
+++ b/distribution/config/nginx.conf
@@ -29,6 +29,8 @@ server {
             # Jenkins environment is booting
             return 521;
         }
+        proxy_set_header Host $host:$server_port;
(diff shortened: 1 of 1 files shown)


distribution/config/nginx.conf
```diff
@@ -29,6 +29,8 @@ server {
             # Jenkins environment is booting
             return 521;
         }
+        proxy_set_header Host $host:$server_port;
+        proxy_redirect http://localhost/ http://localhost:8080/;
```
batmat: `8080` is correct wrt. our current documentation, but is not if user didn't use `-p 8080:80` as documented right?

PR #229, opened 2018-09-06
[JENKINS-53273] JEP-302: initialize Git repo and snapshot state on upgrade

batmat:
Part of [JENKINS-53273](https://issues.jenkins-ci.org/browse/JENKINS-53273) 

**Ready for separate review, and merge if approved.**

The overall story is still WIP, but this PR adds the foundation for the snapshotting:
* the Git/Snapshotter API itself
* integrated with this updates to create commits before upgrades.

I think we could have this merged already separately if the approach looks reasonable. I think it is interesting to deliver this progressively so that we can possibly notice issues already around creating that repository on the disk.

In the meantime, I'm already working on providing the full story based on this: as per JEP-302, leverage this data to be able to roll back if the Health checking tells us we should stop and go back to where we were.

### DIFF
--- a/distribution/client/src/lib/snapshotter.js
+++ b/distribution/client/src/lib/snapshotter.js
@@ -79,7 +79,7 @@ class Snapshotter {
       logger.info(`${LOG_PREFIX} .gitignore outdated or absent, updating.`);
       fs.writeFileSync(gitignorePath, GITIGNORE_CONTENT);
       this.git('add', '.gitignore');
-      this.git('commit', '--message', 'Update.gitignore content to latest');
+      this.git('commit', '--message', 'Update .gitignore content to latest');
     } else {
       logger.info(`${LOG_PREFIX} .gitignore up to date already.`);
     }
--- a/distribution/client/src/lib/snapshotter.js
+++ b/distribution/client/src/lib/snapshotter.js
@@ -1,0 +1,127 @@
+/*
+* Used for snapshotting and rollbacks during upgrades.
+* See:
+*  https://github.com/jenkinsci/jep/tree/master/jep/302
+*/
+
+const { spawnSync } = require('child_process');
+const logger        = require('winston');
+const fs            = require('fs');
+
+const LOG_PREFIX = '[snapshotting]';
+// https://github.com/jenkinsci/jep/tree/master/jep/302#user-content-files-to-store
+const GITIGNORE_CONTENT = `
+/plugins/
+/updates/
+/secrets/master.key
+`;
+
+class Snapshotter {
+
+  constructor(app, options) {
+    this.options = options || {};
+  }
+
+  init(workingDirectory) {
+    if (!workingDirectory) {
+      throw new Error('workingDirectory parameter is required');
+    }
+    this.workingDirectory = workingDirectory;
+
+    if (!fs.existsSync(this.getDotGitDirectory())) {
+      this.git('init');
+      this.git('config', 'user.email', 'evergreen-client@jenkins.io');
+      this.git('config', 'user.name', '"Evergreen Client Snapshotting System (JEP 302)"');
+      this.git('commit','--allow-empty', '--message', 'Initial evergreen commit');
+
+      fs.writeFile(`${this.getDotGitDirectory()}/description`,
+        `This repository is used and specifically designed for Evergreen snapshotting system
+    (https://github.com/jenkinsci/jep/tree/master/jep/302).
+    It is *not* intended to act as a backup system in any way.
+    Never push this repository outside of the Evergreen container to avoid any risk of
+    leaking sensitive data.`,
+        (err) => {
+          if (err) {
+            logger.error('Impossible to write the repository description. Abnormal but not blocking.', err);
+          }
+        });
+    } else {
+      logger.info(`${LOG_PREFIX} ${this.getDotGitDirectory()} already initialized, moving on`);
+    }
+    this.updateGitIgnore();
+  }
+
+  /**
+   * https://github.com/jenkinsci/jep/tree/master/jep/302#take-a-snapshot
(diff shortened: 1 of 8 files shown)


distribution/client/src/lib/snapshotter.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+    this.workingDirectory = workingDirectory;
+
+    if (!fs.existsSync(this.getDotGitDirectory())) {
+      this.git('init');
+      this.git('config', 'user.email', 'evergreen-client@jenkins.io');
+      this.git('config', 'user.name', '"Evergreen Client Snapshotting System (JEP 302)"');
+      this.git('commit','--allow-empty', '--message', 'Initial evergreen commit');
+
+      fs.writeFile(`${this.getDotGitDirectory()}/description`,
```
rtyler: Is this intentionally async? Just curious

distribution/client/src/lib/snapshotter.js
```diff
@@ -0,0 +1,127 @@
...
+    It is *not* intended to act as a backup system in any way.
+    Never push this repository outside of the Evergreen container to avoid any risk of
+    leaking sensitive data.`,
+        (err) => {
+          if (err) {
+            logger.error('Impossible to write the repository description. Abnormal but not blocking.', err);
+          }
+        });
+    } else {
+      logger.info(`${LOG_PREFIX} ${this.getDotGitDirectory()} already initialized, moving on`);
```
rtyler: please lower to `debug`

distribution/client/src/lib/snapshotter.js
```diff
@@ -0,0 +1,127 @@
...
+    this.git('add', '--all');
+    // TODO: allow-empty but should trigger a warning if nothing was changed
+    this.git('commit', '--allow-empty', '--message', `${message}`);
+  }
+
+  /**
+   * Will update the .gitignore file and commit if needed.
+   * No-Op if already containing the expected content.
+   */
+  updateGitIgnore() {
```
rtyler: I'm wondering if the `.gitignore` file should simply be part of the distribution that we ship? Do you think it will change over time?

rtyler: I'm wondering if the `.gitignore` file should simply be part of the distribution that we ship? Do you think it will change over time?

distribution/client/src/lib/update.js
```diff
@@ -88,6 +94,7 @@ class Update {
     });
 
     return Promise.all(tasks).then(() => {
+      this.snapshotter.snapshot(`UL{this.getCurrentLevel()}->UL${updates.meta.level} Snapshot after downloads completed, before Jenkins restart`);
```
rtyler: Please use `UI.publish` a message here about snapshotting the state, that will be exposed to the nginx status page

distribution/client/src/lib/snapshotter.js
batmat: Well, I did give this some thought, and this looked easier for the same outcome to put this in the code directly. As we're able to update the js client code anyway, this looked the same to me as referring to an updatable `.gitignore` file from the disk.

distribution/client/src/lib/snapshotter.js
batmat: Yes, see the message below for the error processing. I just do not want to stop anything just for this in case this goes wrong.

distribution/client/src/lib/snapshotter.js
batmat: Right. Can do, though I did put a bunch of logs, granted, but with the thought that 1) it's going to be interesting to have more than less during some battle testing period and 2) given we're currently setting up what's needed to update the client code, we can update this in production once we know it's working fine.

Also, normally this log should even be called once per client startup, hence normally just when the whole Docker container does.

batmat: done

distribution/client/src/lib/update.js
batmat: done

batmat:
Rebased to resolve the conflict with master.

distribution/client/src/lib/update.js
```diff
@@ -113,7 +121,7 @@ class Update {
   }
 
   saveUpdateSync(manifest) {
-    logger.info('Saving a new manifest..');
+    logger.info('Saving a new manifest.. into ' + this.updatePath());
```
rtyler: In the future please prefer template strings over string concatenation. If i could find an eslint rule for this I would add that :slightly_frowning_face:

distribution/client/test/snapshotter.test.js
```diff
@@ -0,0 +1,48 @@
...
+      const snapshotter = new Snapshotter();
+      snapshotter.init(tmpDir.name);
+
+      // Called twice does not crash:
+      // important since this is what will happen
+      // on container/client restart
+      snapshotter.init(tmpDir.name);
+    });
+  });
+  describe('snapshot()', () => {
```
rtyler: There is no test coverage of the `git()` function which is I believe the most risky one in this module.

We're a bit crunched for time, but @batmat would you please file a follow up ticket to add that test coverage for the potential error branches of `git()` into Milestone 2?

batmat: Done https://issues.jenkins-ci.org/browse/JENKINS-53503

distribution/client/src/lib/update.js
batmat: yeah, right. Will do.

distribution/client/test/snapshotter.test.js
batmat: note that snapshot() calls git and, so does the constructor, so the underlying is tested, but why not.

PR #232, opened 2018-09-06
No plugins for you!

rtyler:

PR #234, opened 2018-09-10
Stop exposing all ports, and do as the docs say currently

batmat:
[JENKINS-53494](https://issues.jenkins-ci.org/browse/JENKINS-53494)

This reverts some redirections introduced previously to get Evergreen back working.
I initially was digging into Nginx docs, but finally went to a simpler revert given our time constraint.

We also get back to exposing a single port from the Evergreen instance to be closer to prod conditions.

Will file a dedicated JIRA to reintroduce this important feature to hide the plugin manager later on (as we do not allow adding plugins by users  at this stage).

### DIFF
--- a/distribution/config/nginx.conf
+++ b/distribution/config/nginx.conf
@@ -111,6 +111,16 @@ http {
             proxy_pass http://jenkins;
         }
 
+        # Compatibility shim to allow Jenkins to have an /evergreen as well as
+        # the root nginx. This should only really be useful for testing, or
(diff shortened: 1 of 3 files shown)


batmat:
Woops, no. I expect `test_manage_plugins_restricted` to still fail. If CI confirms, then I think we're almost there.

batmat:
Note to self, file a JIRA to reintroduce later one some way to hide /pluginManager, which this PR reverts from https://github.com/jenkins-infra/evergreen/pull/232.

PR #235, opened 2018-09-10
But what if redirects actually did the right thing

rtyler:
This will ensure that redirects take the port into account from nginx.


batmat:
Tested locally, this time I couldn't find a way to get stuck :). The fact I cannot access Jenkins root, but am always redirected to _Blue Ocean_ is a bit unsettling for an experienced user, but overall I was still able to do what I needed, so yey!

Thanks!

PR #240, opened 2018-09-11
Embed commit versions into the backend and client containers

rtyler:


services/Makefile
```diff
@@ -9,6 +9,24 @@ endif
...
+# Checking to see if the directory exists because npm install updates the
+# directory every time it runs, busting the GNU/Make cache causing rebuilds
+depends: package.json
+	@if [ ! -d node_modules ]; then \
+		npm install; \
+	fi;
+
+docs: depends
+
+store-commit:
```
batmat: Shouldn't this now go into the `node.mk` file?

rtyler: `distribution/Makefile` doesn't include `node.mk`

batmat:
Was not an approval on purpose, just a comment to ask for the test addition. Not critical but well

rtyler:
I don't think those tests add much value right now.

This is only being added for informational value in the footer of the dashboard, which is not IMO worthy of an acceptance test.

PR #242, opened 2018-09-11
Teach the client how to tell the server about its new update levels

rtyler:
Manually tested, unfortunately we don't have great tooling for the interactions between client and server right now. :slightly_frowning_face:

### DIFF
--- a/distribution/client/src/lib/status.js
+++ b/distribution/client/src/lib/status.js
@@ -68,6 +68,26 @@ class Status {
       });
   }
 
+  reportLevel(updateLevel) {
+    let api = this.app.service('status');
+    let record = {
+      uuid: this.uuid,
+      updateId: updateLevel,
+    };
+    return api.patch(null, record, {
+      query: {
+        uuid: this.uuid,
+      },
+      headers: { Authorization: this.token }
+    })
+      .then(() => {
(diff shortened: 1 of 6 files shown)


distribution/client/src/lib/status.js
```diff
@@ -68,6 +68,26 @@ class Status {
...
+      query: {
+        uuid: this.uuid,
+      },
+      headers: { Authorization: this.token }
+    })
+      .then(() => {
+        logger.info(`Updated the UL on the backend to ${updateLevel}`);
+      })
+      .catch((err) => {
+        logger.error('Failted to update the Status record with the current level', err);
```
mandie722: the extra t is bugging me :)

services/Makefile
```diff
@@ -90,6 +90,13 @@ dump: stop migrate
 		--inserts \
 		-h db -U postgres -d evergreen_development > $(DB_DUMP)
 	$(MAKE) stop
+
+
+seed-ingest:
+	curl --data-raw '{"commit":"$(shell date)","manifest":$(shell cat ../services/ingest.json)}' \
```
batmat: heh, `../services/ingest.json` still works indeed :stuck_out_tongue_closed_eyes:

rtyler: copy paste ftw :smiling_imp:

PR #244, opened 2018-09-12
[JEP-306] Update Level Rollback & Health-checking API

batmat:
The structure of the PR is ready for review, I would even want to have it merged possibly. See the rationale and missing parts below:

TODO (to allow merging)
* [x] Fix the retry logic for healthchecking so that it uses an exponential backoff and does not do *all* the 10 attempts in a few milliseconds like now :man_facepalming: 

TODO (possibly in a followup PR)
* [ ] write/wire the currently empty functions to actually revert UL *and* disk content
* [ ] write an automated test

### DIFF
--- a/distribution/client/src/lib/snapshotter.js
+++ b/distribution/client/src/lib/snapshotter.js
@@ -74,6 +74,9 @@ class Snapshotter {
     this.git('commit', '--allow-empty', '--message', `${message}`);
   }
 
+  revertToLevelBefore(currentLevel) {
+    logger.error(`[NOT IMPLEMENTED YET]${LOG_PREFIX} Revert UL-${currentLevel} to previous Update Level state`);
+  }
   /**
    * Will update the .gitignore file and commit if needed.
    * No-Op if already containing the expected content.
--- a/distribution/client/src/lib/update.js
+++ b/distribution/client/src/lib/update.js
@@ -14,6 +14,7 @@ const Downloader  = require('./downloader');
 const Supervisord = require('./supervisord');
 const UI          = require('./ui');
 const Snapshotter = require('./snapshotter');
+const HealthChecker = require('./healthchecker');
 
 class Update {
   constructor(app, options) {
@@ -26,6 +27,7 @@ class Update {
     this.fileOptions = { encoding: 'utf8' };
     this.snapshotter = new Snapshotter();
     this.snapshotter.init(Storage.jenkinsHome());
+    this.healthChecker = new HealthChecker('http://127.0.0.1:8080'); // FIXME: define a process.env var?
   }
 
   authenticate(uuid, token) {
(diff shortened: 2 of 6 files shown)


batmat:
Timeout. Something is very slow apparently. Problem also seen on JCasC side.

rtyler:
Looking promising thus far!

batmat:
Rebased and force pushed to resolve conflict.

distribution/client/src/lib/snapshotter.js
```diff
@@ -74,6 +74,9 @@ class Snapshotter {
     this.git('commit', '--allow-empty', '--message', `${message}`);
   }
 
+  revertToLevelBefore(currentLevel) {
+    logger.error(`[NOT IMPLEMENTED YET]${LOG_PREFIX} Revert UL-${currentLevel} to previous Update Level state`);
```
batmat: To be written, not the most complex part though. See the UL revert below for the more complex IMO.

distribution/client/src/lib/update.js
```diff
@@ -99,18 +100,89 @@ class Update {
...
+   * 4) if KO, rollback state + go back to previous UL.
+   *
+   * But what if Jenkins does not restart even after rollback?
+     we will not try to go back further. But the client will stay
+
+     Open questions:
+     * how to avoid going again through this UL that borked the system,
+       while it's not been yet marked as tainted in the backend?
+     * how to report that borked case in a clear way
+  */
```
batmat: Note to self: need to remove/rewrite this as it's still in the scratchpad state I left it in since I started thinking about all this.

distribution/client/src/lib/update.js
```diff
@@ -99,18 +100,89 @@ class Update {
...
+          }
+        }
+        Storage.removeBootingFlag();
+      }).catch((errors) => {
+        logger.warn(`TODO ${errors}`);
+      }); // TODO catch?
+  }
+
+  revertToPreviousUpdateLevel() {
+    logger.error(`[NOT IMPLEMENTED YET] Revert UL-${this.getCurrentLevel()} to previous Update Level`);
```
batmat: To be implemented. 

As discussed yesterday with @rtyler, we need to do the following:

* call the to be written new `this.taintCurrentLevel()` function
* then retrigger a `query()`, which should serve the *previous* update.

batmat:
@rtyler I have more in mind to do the wiring with the per-instance tainting API part as a followup. Helps checking it works and seems to work (I'll test/see the logs in "production") before I actually make it do the rollbacks.
But I'm going to merge right away here the exponential backoff healthchecking now allowed by #254 (which, without, would trigger a rollback basically all the time, since last healthchecking was done long before Jenkins had finished starting).

batmat:
Ok, Green. Merging.

PR #245, opened 2018-09-12
Test for JENKINS-53040

rtyler:


rtyler:
FYI, this requires a very production-level test since it will be exercising some of our jenkins-infra/community-functions behaviors

PR #247, opened 2018-09-12
Incorporate Sentry into the backend to make sure we're seeing those errors too

rtyler:
I have already created an evergreen-backend project in Sentry, and will commit
the DSN into our Kubernetes pod shortly

### DIFF
--- a/services/src/app.js
+++ b/services/src/app.js
@@ -31,14 +31,28 @@ const sequelizeSwagger = require('./sequelize-swagger');
 const settings = configuration();
 const app = express(feathers());
 
+
 // const SUCCESS = 'OK';
 const FAILURE = 'ERROR';
 
 /*
-  * Allow the log level to be overridden in the environment for debugging
-  * purposes by the user
-  */
+ * Allow the log level to be overridden in the environment for debugging
+ * purposes by the user
+ */
 logger.level = process.env.LOG_LEVEL || 'warn';
+
+
+/*
+ * Configure Sentry integration only we have a SENTRY_DSN environment
+ */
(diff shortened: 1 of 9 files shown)


rtyler:
The support for this in production will be available once this has been merged https://github.com/jenkins-infra/jenkins-infra/pull/1106

services/src/app.js
```diff
@@ -31,15 +31,29 @@ const sequelizeSwagger = require('./sequelize-swagger');
...
+ */
+if (process.env.SENTRY_DSN) {
+  logger.info('Configuring Sentry for backend errors..');
+  const Raven = require('raven');
+  Raven.config(process.env.SENTRY_DSN).install();
+  app.use(Raven.requestHandler());
+  app.use(Raven.errorHandler());
+  logger.info('..Sentry middleware installed');
+}
+
```
mandie722: This is duplicating the configuration done further down by the sentry class, isn't it?

rtyler: Hm, you're right, damn you singeltons!

services/src/libs/sentry.js
```diff
@@ -34,6 +35,9 @@ class Sentry {
     const errorData = {
       level: data.log.level.toLowerCase(),
       logger: data.log.name,
+      user: {
+        uuid: data.uuid,
```
batmat: :bug: I think, you probably meant to write `id: data.uuid`. https://docs.sentry.io/learn/context#capturing-the-user

I believe that once changed, this will fix https://issues.jenkins-ci.org/browse/JENKINS-53286 right?

PR #251, opened 2018-09-13
Significant refactor of client-server auth code to support JWT re-issuance

rtyler:
Dumping some commit messages here for your convenience

    Refactor server-side authentication to rely more heavily on @feathersjs
    
    While working through my previous commit, it became clear that there was too
    much hacked together code on the server side. With the experience gained on this
    project, I was able to refactor the server to more heavily rely on the
    authentication-local and authentication-jwt modules provided by @feathersjs

---

    Refactor the client to use rely solely on socket.io for improved reliability on authentication errors
    
    This commit removes a lot of manual authentication management for our feathers
    client implementation and instead uses the authentication-client code.
    Unfortunately the authentication-client code doesn't support proper handling of
    re-authentication requirements, and a JWT expiring will simply halt the client
    from communicating with the server.
    
    It ended up being relatively painless to rely soley on the socket.io transport
    for the feathers client which simplifies the footprint a bit and handles
    re-authentication flows.
    
    Much of this change has been manually tested, but will be further tested by our
    smoke tests.
    
    Fixes JENKINS-53543, JENKINS-52525


batmat:
Fuuuu, seems like image deployment is borked again. Last one was 24 hours ago and we do need to update asap :disappointed:

PR #252, opened 2018-09-13
[Jenkins 53257] client deletes

mandie722:
[JENKINS-53257](https://issues.jenkins-ci.org/browse/JENKINS-53257)

Client will delete plugins even if there's no updates

### DIFF
--- a/distribution/client/src/lib/storage.js
+++ b/distribution/client/src/lib/storage.js
@@ -67,12 +67,14 @@ class Storage {
       return;
     }
     let pluginPath = this.pluginsDirectory();
+    let retArray = [];
     plugins.forEach((plugin) => {
-      fs.unlink(`${pluginPath}/${plugin}.hpi`, () => {
+      retArray.push(fs.unlink(`${pluginPath}/${plugin}.hpi`, () => {
         logger.info(`${pluginPath}/${plugin}.hpi was deleted`);
         UI.publish(`Deleted ${plugin}.hpi`);
-      });
+      }));
     });
+    return retArray;
   }
 }
 
--- a/distribution/client/src/lib/storage.js
+++ b/distribution/client/src/lib/storage.js
@@ -1,7 +1,9 @@
 'use strict';
 
-const path = require('path');
-const fs   = require('fs');
+const path   = require('path');
+const fs     = require('fs');
+const logger = require('winston');
+const UI     = require('./ui');
 
 /*
  * The Storage module simply contains common functions necessary for the
@@ -58,6 +60,19 @@ class Storage {
     if (fs.existsSync(filePath)) {
       fs.unlinkSync(filePath);
     }
+  }
+
+  static removePlugins(plugins) {
+    if (!plugins) {
+      return;
+    }
+    let pluginPath = this.pluginsDirectory();
+    plugins.forEach((plugin) => {
+      fs.unlink(`${pluginPath}/${plugin}.hpi`, () => {
+        logger.info(`${pluginPath}/${plugin}.hpi was deleted`);
+        UI.publish(`Deleted ${plugin}.hpi`);
+      });
+    });
   }
 }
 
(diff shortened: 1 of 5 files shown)


distribution/client/test/helpers.js
```diff
@@ -0,0 +1,26 @@
...
+ * tests
+ */
+const fs            = require('fs');
+
+class Helpers {
+  constructor () {
+  }
+
+  checkFileExists(filename) {
+    return new Promise(r=>fs.access(filename, fs.F_OK, e => r(!e)));
```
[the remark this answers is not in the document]
batmat: Why not `fs.existsSync(path)`?
https://nodejs.org/api/fs.html#fs_fs_existssync_path

Also do you really want to be async in that case? I guess not in tests?

distribution/client/src/lib/update.js
```diff
@@ -74,7 +74,7 @@ class Update {
         updates.core.checksum.signature));
     }
 
-    if ((!updates.plugins) || (!updates.plugins.updates)) {
+    if ((!updates.plugins) || (!updates.plugins.updates && !updates.plugins.deletes)) {
       logger.debug('No plugins available in the response');
       this.updateInProgress = null;
       this.saveUpdateSync(updates);
```
batmat: :fire_engine: 
Unrelated with your change, but isn't there a huge bug here? If only the core is updated, we'll bail out without restart nor snapshotting, but the war has been downloaded and put at the location ready isn't it?

distribution/client/src/lib/update.js
```diff
@@ -86,13 +86,19 @@ class Update {
...
+        logger.info('Downloading', plugin);
+        tasks.push(Downloader.download(plugin.url,
+          Storage.pluginsDirectory(),
+          `${plugin.artifactId}.hpi`,
+          plugin.checksum.signature));
+      });
+    }
+
+    if (updates.plugins.deletes) {
+      tasks.push(Storage.removePlugins(updates.plugins.deletes));
```
batmat: Can we add more `logger.info` and the likes inside those `if` conditions? I feel at this stage of the project, we want rather more logs than fewer don't we?

distribution/client/src/lib/storage.js
```diff
@@ -59,6 +61,21 @@ class Storage {
...
+    }
+    let pluginPath = this.pluginsDirectory();
+    let retArray = [];
+    plugins.forEach((plugin) => {
+      retArray.push(fs.unlink(`${pluginPath}/${plugin}.hpi`, () => {
+        logger.info(`${pluginPath}/${plugin}.hpi was deleted`);
+        UI.publish(`Deleted ${plugin}.hpi`);
+      }));
+    });
+    return retArray;
```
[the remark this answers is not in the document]
mandie722: I have tests that confirm this works, but they conflict with the snapshotting changes, so I'm still rewriting to get it all working together.

distribution/client/test/helpers.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  touchFile(filename) {
```
batmat: sync `fs.openSync`, or even `fs.writeFileSync('/path/to/theFile', 'blah content')` might be easier for the purpose?

distribution/client/test/storage.test.js
```diff
@@ -29,4 +42,26 @@ describe('The storage module', () => {
...
+      const pluginPath = Storage.pluginsDirectory();
+      mkdirp.sync(pluginPath);
+      filenames.forEach((filename) => {
+        h.touchFile(`${pluginPath}/${filename}.hpi`);
+        expect(h.checkFileExists(`${pluginPath}/${filename}.hpi`)).resolves.toBeTruthy();
+      });
+      await Storage.removePlugins(filenames);
+      filenames.forEach((filename) => {
+        expect(h.checkFileExists(`${pluginPath}/${filename}.hpi`)).resolves.toBeFalsy();
+      });
```
batmat: nice one

distribution/client/src/lib/update.js
mandie722: true, hadn't found that one yet, we have a severe lack of positive testing that I'm trying to fix

PR #253, opened 2018-09-13
Provide an API for marking instance-specific update levels as tainted

rtyler:
Changes to the update service still forthcoming

### DIFF
--- a/services/migrations/20180913193732-create-tainted.js
+++ b/services/migrations/20180913193732-create-tainted.js
@@ -1,0 +1,44 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('tainteds', {
+      id: {
+        allowNull: false,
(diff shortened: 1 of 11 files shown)


services/migrations/20180913193732-create-tainted.js
```diff
@@ -0,0 +1,36 @@
...
+      uuid: {
+        type: Sequelize.UUID
+      },
+      updateId: {
+        type: Sequelize.INTEGER,
+        references: {
+          model: 'updates',
+          key: 'id'
+        },
+      },
```
batmat: Not a big deal anyway if there's dupes, but I guess we should define a unique composite key. IIUC, this would just be adding like
`unique: 'instance-tainted'` on both `uuid` and `updateId` fields.

rtyler: good suggestion, will add

PR #254, opened 2018-09-14
Set up module patching and use patched promise-request-retry

batmat:
While a few PRs are in flight, we can move forward without needing to go through an even heavier fork process.

Integrates https://github.com/void666/request-promise-retry/pull/6 and https://github.com/void666/request-promise-retry/pull/5 while they're not merged & released yet.

### DIFF
--- a/distribution/client/package.json
+++ b/distribution/client/package.json
@@ -7,7 +7,8 @@
     "client": "node src/client.js",
     "test": "npm run eslint && npm run jest",
     "eslint": "eslint src/. test/. --config .eslintrc.json",
-    "jest": "jest"
+    "jest": "jest",
+    "prepare": "patch-package"
   },
   "author": "R Tyler Croy",
   "license": "GPL-3.0",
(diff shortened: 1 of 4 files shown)


distribution/client/package.json
```diff
@@ -7,7 +7,8 @@
     "client": "node src/client.js",
     "test": "npm run eslint && npm run jest",
     "eslint": "eslint src/. test/. --config .eslintrc.json",
-    "jest": "jest"
+    "jest": "jest",
+    "prepare": "patch-package"
```
rtyler: This should be executed in the `Makefile` shouldn't it?

batmat:
It's done automatically.  Because the patching is bound to the standard
'prepare' script.

Le ven. 14 sept. 2018 à 17:59, R. Tyler Croy <notifications@github.com> a
écrit :

distribution/client/package.json
batmat: It's done automatically. Because the patching is bound to the standard 'prepare' script. 

BTW that is also why I added the associated test: to check and show this does get applied.

PR #256, opened 2018-09-14
Downgrade to @feathersjs/feathers 3.1.7

rtyler:
See https://github.com/feathersjs/feathers/issues/1000


batmat:
@rtyler so for the record, did you do what David recommended in the linked PR, that is deleting the package.json and recreate from scratch to get npm to choose all the right versions?

PR #260, opened 2018-09-15
Consider the instance's flavor when comparing its versions to the upd…

rtyler:
…ate level

I believe this fixes JENKINS-53318, but I noticed this when my demo instance had
its docker-plugin, ssh-slaves, and docker-java-api plugins deleted. It had
reported the versions properly when it first installed
(https://gist.github.com/rtyler/1ee1cfc5f318a02bcbf47c59c376a384) but subsequent
update requests caused the deletion of those plugins thereby breaking the
instance.

This commit ensures that the flavor is considered when comparing versions to
also compare the versions of plugins provided by the flavor

Fixes JENKINS-53318

### DIFF
--- a/services/acceptance/update-versions.test.js
+++ b/services/acceptance/update-versions.test.js
@@ -5,7 +5,6 @@
  */
 
 const fs      = require('fs');
-const logger  = require('winston');
 const request = require('request-promise');
 const h       = require('./helpers');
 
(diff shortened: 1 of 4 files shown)


rtyler:
@batmat is on an airplane and @mandie722 is hopefully enjoying her weekend. I think this is safe enough to merge and deploy to get into production

PR #269, opened 2018-09-17
Relocate data into a volume so docker pull jenkins/evergreen:docker-cloud will work

rtyler:


distribution/client/src/lib/storage.js
```diff
@@ -19,10 +19,10 @@ class Storage {
     /* The default home directory is /evergreen, see the Dockerfile in the root
      * directory of th repository
      */
-    if (!process.env.EVERGREEN_HOME) {
-      return '/evergreen';
+    if (!process.env.EVERGREEN_DATA) {
```
batmat: Not sure about this one. It's going to be confusing in the code. Maybe better to introduce a new utility function here or so? `jenkinsHome()` returning data, not home basically?

distribution/config/nginx.conf
```diff
@@ -122,7 +122,7 @@ http {
         }
 
         location / {
-            if ( -f /evergreen/booting.txt ) {
+            if ( -f /evergreen/data/booting.txt ) {
```
batmat: Typically shouldn't it be left above on purpose? I mean: we should put things in `$EVERGREEN_DATA` we care about to keep for long. AFAICT, we do not care about the booting.txt file staying over a `docker stop` and new pull do we?

distribution/client/src/lib/storage.js
rtyler: `jenkinsHome()` already exist one function below this :smile:

distribution/config/nginx.conf
rtyler: I thought about that too. This keeps things a bit simpler in the client in that it will not _write_ anything outside of `EVERGREEN_DATA` which is IMHO a good thing.

PR #284, opened 2018-09-24
Checksum docker binaries download and remove wget

batmat:
I added wget back in the days to add the shim for downloading plugins.

Now we have the client download, I do not think it's needed anymore and we have `curl` that can download things too anyway.
Removing wget will reduce a wee bit the download size for the layer.

### DIFF
--- a/distribution/Dockerfile
+++ b/distribution/Dockerfile
@@ -60,9 +60,7 @@ RUN apk add --no-cache git \
                         ttf-dejavu \
                         curl \
                         socat \
-                        wget \
-                        nginx \
-                        git # JEP-302
+                        nginx
 
 # Ensure the latest npm is available
 RUN npm install npm@latest -g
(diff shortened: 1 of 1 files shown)


distribution/Dockerfile
```diff
@@ -60,20 +60,18 @@ RUN apk add --no-cache git \
                         ttf-dejavu \
                         curl \
                         socat \
-                        wget \
-                        nginx \
-                        git # JEP-302
```
batmat: Also removed `git`, because it was already specified on the L53 above :man_facepalming:.

batmat:
Unrelated failure, AFAICT, but for the record I found that the parameters I'm passing in are actually ignored. 
image

So I'm fixing the test(s).

batmat:
Thing is: there's no obvious solution/fix either I guess:

> info: Download complete for /tmp/ace-editor.hpi (Took 128.903s)

I.e. the download worked, but took more than the 120s timeout.
Now, is this normal that downloading a 5 MB file from Azure to (some?) mirror can take 128 seconds, i.e. be 40 kB/sec, is another question.

batmat:
Proposed https://github.com/jenkins-infra/evergreen/pull/285 to try to improve the statu quo.

PR #285, opened 2018-09-24
Deflake a bit

batmat:
I've fixed/improved a few places. Put a dedicated commit each time to keep track of the rationale for each change.

### DIFF
--- a/distribution/client/test/downloader.test.js
+++ b/distribution/client/test/downloader.test.js
@@ -14,6 +14,9 @@ describe('the Downloader class', () => {
       /* Make sure memfs is flushed every time */
       fs.volume.reset();
       mkdirp.sync(dir);
+
+      // default is 5 seconds, could be bigger because of the various file sizes and network flakiness
+      jest.setTimeout(120000);
     });
 
     it('should return promise', async () => {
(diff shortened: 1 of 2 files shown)


distribution/client/test/downloader.test.js
```diff
@@ -50,21 +49,21 @@ describe('the Downloader class', () => {
...
-        const toDownload = 'http://nonexisting-url-yada.org/thefile';
+        const toDownload = 'http://nonexisting-url-dont-retry-too-fast.org/thefile';
         const startTime = new Date();
         try {
-          await Downloader.download(toDownload, dir, 'thefile', {retry: 4, factor: 1.1, delay: 200});
+          await Downloader.download(toDownload, dir, 'thefile', null, {delay: 20, retry: 4, factor: 10});
           expect(false).toBeTruthy(); // fail(), should not reach this line.
         } catch (e) {
-          expect(new Date() - startTime).toBeGreaterThan(4 * 200);
+          expect(new Date() - startTime).toBeGreaterThan(0 + 20 + 200 + 2000);
```
rtyler: This little equation merits a code comment explaining what this magic number is for, I'm a bit confused looking at it myself.

mandie722: I agree on needing the explanation for this math, also the reason for the values on delay, retry, and factor would be awesome too.

PR #288, opened 2018-09-24
Re-organize docs and fix message wrapping in the UI

rtyler:
Bringing docs up into the root directory for easier contribution.


distribution/client/public/index.html
```diff
@@ -30,6 +30,9 @@
             .connected {
                 background-color: #006a27;
             }
+            pre {
+                white-space: pre-wrap;
+            }
```
batmat: I'm inclined to think that this might not be the right fix: the `index.js` inserts elements using `document.createElement('pre')`, when these are actually messages we probably want wrapping and default formatting to act on, instead of preformatted code (like program source code).
I.e. inserting there `<p>` for instance, and use css to format like we would like to appear.

PR #293, opened 2018-09-26
Re-locate the Sentry error handler to take effect.

rtyler:
Apparently installing this errorHandler() middleware too early results in
feathers errors not being properly caught and reported.

I've verified this by running acceptance tests and seeing what gets reported in
Sentry (lots! :smile:)

Fixes JENKINS-53749

### DIFF
--- a/services/src/app.js
+++ b/services/src/app.js
@@ -7,6 +7,7 @@ const compress       = require('compression');
 const cors           = require('cors');
 const helmet         = require('helmet');
 const logger         = require('winston');
+const Raven          = require('raven');
 
 const feathers       = require('@feathersjs/feathers');
 const configuration  = require('@feathersjs/configuration');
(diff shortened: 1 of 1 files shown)


services/src/app.js
```diff
@@ -49,10 +50,8 @@ logger.level = process.env.LOG_LEVEL || 'warn';
  */
 if (process.env.SENTRY_DSN) {
   logger.info('Configuring Sentry for backend errors..');
-  const Raven = require('raven');
   Raven.config(process.env.SENTRY_DSN).install();
   app.use(Raven.requestHandler());
-  app.use(Raven.errorHandler());
   logger.info('..Sentry middleware installed');
 }
```
mandie722: Is there a reason the rest of the logic wouldn't also move to the end of app startup?  It is better for code readability if all the configuration for Sentry is all together.

rtyler: Yeah, the `requestHandler` has  to be the first middleware :man_facepalming: 

It's derpy

mandie722: Super derpy

rtyler:
@mandie722 I'm going to add some comments explaining why this is broken up a bit to avoid confusion in the future, then I'll merge

PR #296, opened 2018-09-28
Add an unused System Property which can be used to determine if we're running in Evergreen

rtyler:
I expect that this will be useful further down the line for toggling specific
plugin behaviors on whether they're in the Evergreen distribution or not

### DIFF
--- a/distribution/Dockerfile
+++ b/distribution/Dockerfile
@@ -20,6 +20,7 @@ ENV JENKINS_UC https://updates.jenkins.io
 ENV JENKINS_UC_EXPERIMENTAL=https://updates.jenkins.io/experimental
 
 ENV JAVA_OPTS=\
+"-Djenkins.evergreen=true "\
 "-Djava.awt.headless=true "\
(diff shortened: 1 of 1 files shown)


batmat:
Well, why not, but it seems to me we can already do that using `EVERGREEN_HOME` env var or any other can't we?

rtyler:
@batmat I was thinking that the system property might be useful for sneaky people using "regular old Jenkins" who wanted to turn on specific functionality in plugins which we might gate with this system property feature flag.

That said, if a plugin other than the Evergreen plugin needed to check "am I running in Evergreen?" checking environment variables seems kind of like a hacky way to do it IMO. I would rather support this system property as our "API" rather than environment variables which are pretty tied to our current implementation outside of Jenkins.

rtyler:
I had some thoughts on Pipeline and some other plugins adding functionality which would be toggled on automatically. But without something in our users hands that tells Pipeline that its running on Evergreen, those flagged features would be pointless.


I give up on debating this.

PR #297, opened 2018-10-01
[JENKINS-53856] Fix 'Permission denied' errors

batmat:
[JENKINS-53856](https://issues.jenkins-ci.org/browse/JENKINS-53856)

* [x] Add test surfacing this
* [x] Actually fix the issue


rtyler:
Is this the `npm` related log spam? I think that might just be the `HOME` directory is not set properly for the user that runs out of supervisord. :man_shrugging:

batmat:
@rtyler not only. Git too. But yes, that's it.

batmat:
Also verified locally to fix the issue. Now waiting for a green confirmation. :)

PR #300, opened 2018-10-03
[WIP] Automated rollback

batmat:


distribution/client/test/update.test.js
```diff
@@ -26,6 +25,14 @@ describe('The update module', () => {
     app = feathers();
     update = new Update(app);
     update.healthChecker = new HealthChecker('http://127.0.0.1:8080/', { delay: 50, retry: 5});
```
batmat: I think this line could be simply a `new Object()`, but well, that shouldn't be biggy different.

batmat:
Just to clarify my intent. I'm not just hoping that thing will go to green randomly here at some point and will merge immediately then. 
I'm retriggering because I'm each time checking if this behaves/fails the same way, and it's not the case. 

I hope/expect #304 should make the failure and the way it fails much more regular/always the same.

distribution/client/test/update.rollback.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      // query cannot really call the backend side, so we mock it to return ul1
+      update.query = jest.fn( () => {
+        return Promise.resolve(ul1);
+      });
+
+
+      expect(update.taintUpdateLevel).not.toHaveBeenCalled(); // not yet, let's just double-check
+
+      // update to broken UL2 (as healthchecking is stubbed above to fail)
+      const result2 = await update.applyUpdates(ul2);
```
batmat: here, I would need to test an upgrade actually occurred to 2. Probably just need to use the `mock` below to see what got passed as an UL to revert.

batmat:
Going to file a new one with all this, to avoid the big number of messages like posted above. Stay tuned.

PR #304, opened 2018-10-05
[JENKINS-53499] Improve tests to avoid race condition for updates during startup

batmat:
[JENKINS-53499](https://issues.jenkins-ci.org/browse/JENKINS-53499)

The change here aims at making sure the client will start only when:
* the backend has started, and
* it has already received a first UL

So, we do the following:
* delay the client startup by a few seconds, only in development mode
  (by adding a startup wrapper script that will handle this)
* the test framework tries to upload the UL every seconds, instead of
  every N seconds, with N growing by one on each failed attempt.

### DIFF
--- a/distribution/scripts/start-client.sh
+++ b/distribution/scripts/start-client.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+# Wrapper to force the client to wait for the backend to become available
+# Should only be useful during development
+set -euo pipefail
+
+sleepTime=8
+if [ ! -z ${DEVELOPMENT:-} ]; then
+  echo "DEVELOPMENT MODE: client will wait $sleepTime before starting, to give time to the backend to start and receive a first UL"
+  sleep $sleepTime
+else
+  echo "Client is starting up"
+fi
+
+maxAttempts=30
+
+until curl -s $EVERGREEN_ENDPOINT --output /dev/null ; do
+  maxAttempts=$(( $maxAttempts - 1 ))
+  if [[ $maxAttempts <= 0 ]]; then
+    >&2 echo "Maximum number of attempts reached: exiting"
+    exit 1
+  fi
+  >&2 echo "Backend is unavailable - sleeping for some more time"
+  sleep 1
+done
+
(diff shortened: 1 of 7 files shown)


distribution/scripts/start-client.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Wrapper to force the client to wait for the backend to become available
+# Should only be useful during development
+set -euo pipefail
+
+sleepTime=8
```
rtyler: Not sure why this variable is necessary :man_shrugging:

distribution/scripts/start-client.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ ! -z "${DEVELOPMENT:-}" ]; then
+  echo "DEVELOPMENT MODE: client will wait $sleepTime before starting, to give time to the backend to start and receive a first UL"
+  sleep $sleepTime
+else
+  echo "Client is starting up"
+fi
+
+maxAttempts=30
+
+until curl -s "$EVERGREEN_ENDPOINT" --output /dev/null ; do
```
rtyler: Is this expected to be running in non-development mode as well? In released instances, the client will exit if it cannot reach the backend and allow supervisord to restart it.

Is this change of behavior intentional for those released instances? If so, would you mind adding some comments to the script as to why this is necessary.

batmat: There are already a pretty commit log with this but ok

distribution/scripts/start-client.sh
batmat: True, it could defined in the `if` scope below, if that is what you mean.

But it's needed since I use it in both the message and the `sleep $sleepTime` below on line 10.

distribution/scripts/start-client.sh
batmat: OK, I think I get your point now. Yes, I could, probably should, move that block inside the `if` above too. Will do.

batmat: fixed with latest pushed commit.

PR #305, opened 2018-10-06
Convert evergreen-client to TypeScript

rtyler:
This should make development a bit easier.


rtyler:
Still need to figure out how to best package this up in the container

distribution/client/src/client.ts
```diff
@@ -151,7 +160,7 @@ class Client {
       this.runUpdates();
     });
 
-    this.reg.register().then((res, newRegistration) => {
+    this.reg.register().then(({res, newRegistration}) => {
```
batmat: this arrow function looks fishy?

distribution/client/src/lib/error-telemetry.ts
```diff
@@ -5,13 +5,24 @@
...
-const Storage  = require('./storage');
+import { Tail } from 'tail';
+import fs from 'fs';
+import * as logger from 'winston'
+import path from 'path';
+
+import Storage from './storage';
+
+export interface ErrorTelemetryOptions {
+  flavor?: string,
```
batmat: as I seem to understand, I _think_ this should actually be non optional, but that's not critical anyway and fixable later since the rest of the code makes it crash anyway.

distribution/client/src/lib/status.ts
```diff
@@ -1,20 +1,30 @@
...
-const Storage  = require('./storage');
-const Checksum = require('./checksum');
+import * as logger from 'winston';
+import Storage from './storage';
+import Checksum from './checksum';
 
 
+export interface StatusOptions {
+  flavor?: string,
+};
```
batmat: Seems like there could (should) be some something central for the flavors, as this code is the same as above. (just a comment, not a request for fix here)

distribution/client/src/lib/status.ts
```diff
@@ -38,8 +48,8 @@ class Status {
    * Create a status record in the backend for this particular instance
    */
   async create() {
-    let api = this.app.service('status');
-    let record = {
+    const api = this.app.service('status');
```
batmat: AIUI, these `let` => `const` are unrelated to the typescript migration, and were done incidentally while looking at this code. Or does typescript enforce this? (I'll check/discover that later anyway :))

distribution/client/src/lib/supervisord.ts
```diff
@@ -68,13 +71,11 @@ class Supervisord {
     });
   }
 
-  async restartProcess(name) {
+  static async restartProcess(name) {
```
batmat: same here, I suppose, unrelated to TS migration (?)

distribution/client/src/lib/status.ts
rtyler: Just cleaning up while I was here :smile:

distribution/client/src/client.ts
rtyler: This was actually something caught by the TypeScript conversion. From what I was reading on the interwebs, Promise resolve/reject functions are intended on only having a single value passed through.

Now it's passing a single object through.

distribution/client/src/lib/status.ts
rtyler: I agree, I think there's some refactoring we can do in the client to reduce some redundancies  :+1:

PR #312, opened 2018-10-08
Updated plugins from experimental UC

jglick:
Pulls in JEP-210 beta releases. @svanoort @dwnusbaum

### DIFF
--- a/services/Makefile
+++ b/services/Makefile
@@ -127,8 +127,14 @@ publish: container
 update-center.json:
 	wget -O update-center.json https://updates.jenkins.io/update-center.actual.json
 
+update-center-experimental.json:
+	wget -O update-center-experimental.json https://updates.jenkins.io/experimental/update-center.actual.json
+
 propose-updates: update-center.json
-	node ./prepare-essentials propose-updates
+	node ./prepare-essentials propose-updates --uc ./update-center.json
+
+propose-experimental-updates: update-center-experimental.json
+	node ./prepare-essentials propose-updates --uc ./update-center-experimental.json
 
 generate-essentials: essentials.yaml prepare-essentials update-center.json
 	node ./prepare-essentials save
--- a/services/cli/manifest-resolver.js
+++ b/services/cli/manifest-resolver.js
@@ -85,8 +85,15 @@ class ManifestResolver {
             plugin.optional = false;
           }
 
+          var ok;
+          try {
+            ok = compareVersions(needed[artifactId].version, plugin.version) == 1;
+          } catch (x) {
+            logger.warn(`Cannot compare ${artifactId}:${needed[artifactId].version} to ${plugin.version}: ${x}`);
+            ok = true;
+          }
           // The plugin version requested is lower than one we already have.
-          if (compareVersions(needed[artifactId].version, plugin.version) == 1) {
+          if (ok) {
             if (!plugin.optional) {
               /*
               * If version of the plugin which is already present is considered
@@ -204,6 +211,7 @@ class ManifestResolver {
             return false;
           }).forEach((dep) => {
             envNeeds[dep] = deps[dep];
+            // TODO this is printing bogus messages like: The docker-cloud environment also needs docker-plugin
             logger.info(`The ${environment} environment also needs ${dep}`);
           });
           this.environmentNeeded[environment] = envNeeds;
--- a/services/cli/manifest-resolver.js
+++ b/services/cli/manifest-resolver.js
@@ -85,7 +85,7 @@ class ManifestResolver {
             plugin.optional = false;
           }
 
-          var ok;
+          let ok;
           try {
             ok = compareVersions(needed[artifactId].version, plugin.version) == 1;
           } catch (x) {
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -8,7 +8,7 @@ spec:
   plugins:
     - groupId: io.jenkins
       artifactId: configuration-as-code
-      version: '1.0'
+      version: '1.1'
     - groupId: io.jenkins.plugins
       artifactId: evergreen
       version: 1.0-rc58.a9ecd200b39a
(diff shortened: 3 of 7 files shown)


services/Makefile
```diff
@@ -127,8 +127,14 @@ publish: container
...
 	wget -O update-center.json https://updates.jenkins.io/update-center.actual.json
 
+update-center-experimental.json:
+	wget -O update-center-experimental.json https://updates.jenkins.io/experimental/update-center.actual.json
+
 propose-updates: update-center.json
-	node ./prepare-essentials propose-updates
+	node ./prepare-essentials propose-updates --uc ./update-center.json
+
+propose-experimental-updates: update-center-experimental.json
```
jglick: Just offering an alternate target. Really this stuff should of course be replaced with usage of the updater logic from `incrementals-tools` so that we get a single consistent way to update everything to whatever is in `master`.

services/cli/manifest-resolver.js
```diff
@@ -85,8 +85,15 @@ class ManifestResolver {
             plugin.optional = false;
           }
 
+          var ok;
+          try {
+            ok = compareVersions(needed[artifactId].version, plugin.version) == 1;
+          } catch (x) {
+            logger.warn(`Cannot compare ${artifactId}:${needed[artifactId].version} to ${plugin.version}: ${x}`);
```
jglick: Whatever Node library you are using to compare versions, it is hopeless in the face of the version numbers actually used in Jenkins components. We want to be using `jenkinsci/lib-version-number` instead. (As, for example, `incrementals-tools` does.)

services/cli/manifest-resolver.js
```diff
@@ -204,6 +211,7 @@ class ManifestResolver {
             return false;
           }).forEach((dep) => {
             envNeeds[dep] = deps[dep];
+            // TODO this is printing bogus messages like: The docker-cloud environment also needs docker-plugin
```
jglick: Not sure what this is all about.

services/prepare-essentials
```diff
@@ -13,15 +13,22 @@ logger.level = process.env.LOG_LEVEL || 'info';
...
 yargs.command('propose-updates',
   'Fetch the latest Update Center and propose updates',
-  () => {},
+  {uc: {}},
   (argv) => {
-    const uc = UpdateCenter.fromFile('./update-center.json')
+    const uc = UpdateCenter.fromFile(argv.uc)
     const manifest = Manifest.loadFile();
-    manifest.getPlugins().forEach((plugin) => {
+    manifest.getActualPlugins().forEach((plugin) => {
```
jglick: The updater was failing to show updates to what was actually being delivered in `status`! It only listed newer versions of plugins which were _already_ in `spec`. As I have noted before, this is not helpful, since for example `workflow-aggregator` itself does nothing and should have been removed as of Jenkins 2; we need to actually enumerate the plugins we want to ship.

services/prepare-essentials
```diff
@@ -13,15 +13,22 @@ logger.level = process.env.LOG_LEVEL || 'info';
...
       const updated = uc.plugins[plugin.artifactId]
-      if ((!plugin.version.match(/(.*?)-rc(\d+)\.(.*)?/)) &&
-        (compareVersions(updated.version, plugin.version) == 1)) {
-        logger.info(`The update center has a newer ${plugin.artifactId}: ${updated.version}`);
+      if (updated == null) {
+        logger.warn(`No such plugin ${plugin.artifactId}`);
```
jglick: Hit this at one point, not sure why.

services/essentials.yaml
```diff
@@ -39,27 +45,57 @@ spec:
...
+      version: '2.12'
+    - groupId: org.jenkins-ci.plugins.workflow
+      artifactId: workflow-job
+      version: 2.26-beta-1
+    - groupId: org.jenkins-ci.plugins.workflow
+      artifactId: workflow-scm-step
+      version: '2.7'
+    - groupId: org.jenkins-ci.plugins.workflow
+      artifactId: workflow-support
+      version: 2.21-beta-1
```
jglick: This block is what I was actually trying to accomplish: pull in [JEP-210](https://jenkins.io/jep/210).

services/essentials.yaml
```diff
@@ -8,7 +8,7 @@ spec:
   plugins:
     - groupId: io.jenkins
       artifactId: configuration-as-code
-      version: '1.0'
+      version: '1.1'
```
jglick: Just updating everything while I am here.

services/essentials.yaml
```diff
@@ -39,27 +45,57 @@ spec:
...
+    - groupId: org.jvnet.hudson.plugins
+      artifactId: favorite
+      version: 2.3.2
     - groupId: com.coravy.hudson.plugins.github
       artifactId: github
-      version: 1.29.2
+      version: 1.29.3
+    - groupId: org.jenkins-ci.plugins
+      artifactId: github-api
+      version: '1.92'
```
jglick: Seems the YAML serializer is more clever than we might like: `x.y` is quoted as a string but `x.y.z` is not, even though for our purposes we know all `version`s ought to be loaded as strings. JavaScript :man_shrugging:

services/essentials.yaml
```diff
@@ -87,6 +129,15 @@ spec:
     - groupId: org.jenkins-ci.plugins
       artifactId: jira
       version: 3.0.2
+    - groupId: org.jenkins-ci.ui
+      artifactId: ace-editor
```
jglick: Hmm, we sort the `status` when we write it out; should we not also sort the `spec`?

(My advice remains to get rid of the separate `spec` and `status` sections.)

services/cli/manifest-resolver.js
batmat: I BTW, https://issues.jenkins-ci.org/browse/JENKINS-53767 is probably worth looking at. I agree at some point we'd need a central tool to do all this.

jglick:
> support experimental UC in addition to the main one, though I wonder why we shouldn't anyway just jump to just using Incrementals

That would be my preference. I was making the minimal change I needed to get this particular set of updates in (with repeatable tooling rather than one-off edits). As noted in https://github.com/jenkins-infra/evergreen/pull/312#discussion_r223473232 I am not aware of _existing_ tooling to do this.

jglick:
> cases where we'd like to consume experimental releases from non-incrementalified plugins

Well, you could always just file a PR to

```sh
mvn incrementals:incrementalify
```

and ship the result. :-)

rtyler:
Going to bring these in before some other updates just to keep the pipes moving

PR #313, opened 2018-10-09
[JENKINS-53964] Drop uuid<->checksum unique constraint from versions table

batmat:
https://issues.jenkins-ci.org/browse/JENKINS-53964

I manually verified that the constraint is not here anymore with this change. 
With that fix:

```
evergreen_development=# \d+ versions;
                                                                Table "public.versions"
        Column         |           Type           | Collation | Nullable |               Default                | Storage  | Stats target | Description
-----------------------+--------------------------+-----------+----------+--------------------------------------+----------+--------------+-------------
 id                    | integer                  |           | not null | nextval('versions_id_seq'::regclass) | plain    |              |
 uuid                  | character varying(255)   |           | not null |                                      | extended |              |
 manifest              | json                     |           |          |                                      | extended |              |
 manifestSchemaVersion | integer                  |           |          |                                      | plain    |              |
 createdAt             | timestamp with time zone |           | not null |                                      | plain    |              |
 updatedAt             | timestamp with time zone |           | not null |                                      | plain    |              |
 checksum              | character varying(255)   |           | not null |                                      | extended |              |
Indexes:
    "versions_pkey" PRIMARY KEY, btree (id)

evergreen_development=# ^D\q
```

Before

```
evergreen_development=# \d+ versions;
                                                                Table "public.versions"
        Column         |           Type           | Collation | Nullable |               Default                | Storage  | Stats target | Description 
-----------------------+--------------------------+-----------+----------+--------------------------------------+----------+--------------+-------------
 id                    | integer                  |           | not null | nextval('versions_id_seq'::regclass) | plain    |              | 
 uuid                  | character varying(255)   |           | not null |                                      | extended |              | 
 manifest              | json                     |           |          |                                      | extended |              | 
 manifestSchemaVersion | integer                  |           |          |                                      | plain    |              | 
 createdAt             | timestamp with time zone |           | not null |                                      | plain    |              | 
 updatedAt             | timestamp with time zone |           | not null |                                      | plain    |              | 
 checksum              | character varying(255)   |           | not null |                                      | extended |              | 
Indexes:
    "versions_pkey" PRIMARY KEY, btree (id)
    "uuid_checksum_uniq" UNIQUE CONSTRAINT, btree (uuid, checksum)
```

### DIFF
--- a/services/acceptance/services/version.test.js
+++ b/services/acceptance/services/version.test.js
@@ -85,15 +85,27 @@ describe('Versions service acceptance tests', () => {
       });
 
       it('should allow creating redundant versions records', async () => {
-        let req = {
+        const req = {
           url: h.getUrl('/versions'),
           method: 'POST',
           headers: { 'Authorization': this.token },
           json: true,
           body: { uuid: this.reg.uuid, manifest: manifest }
         };
-        await request(req);
-        await request(req);
+        let id = null;
+        {
+          const result = await request(req);
+          expect(result.uuid).toBeTruthy();
+          expect(result.uuid).toEqual(this.reg.uuid);
+          expect(id).not.toEqual(result.id);
+          id = result.id;
+        }
+        {
+          const result = await request(req);
+          expect(result.uuid).toBeTruthy();
+          expect(result.uuid).toEqual(this.reg.uuid);
+          expect(id).not.toEqual(result.id); // show it's a new row in db, not an update or so
+        }
       });
     });
   });
--- a/services/acceptance/services/version.test.js
+++ b/services/acceptance/services/version.test.js
@@ -84,7 +84,7 @@ describe('Versions service acceptance tests', () => {
           .catch(err => expect(err.statusCode).toEqual(401));
       });
 
-      it('should disallow creating redundant versions records', async () => {
+      it('should allow creating redundant versions records', async () => {
         let req = {
           url: h.getUrl('/versions'),
           method: 'POST',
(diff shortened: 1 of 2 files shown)


services/migrations/20181009204000-drop-uuid_checksum_uniq-key.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.removeConstraint('versions', 'uuid_checksum_uniq');
+  },
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.addConstraint('versions',
```
batmat: this is basically a copy paste of what's in the versions table creation.

services/migrations/20181009204000-drop-uuid_checksum_uniq-key.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.removeConstraint('versions', 'uuid_checksum_uniq');
```
batmat: **this** is the gist of this change.

services/acceptance/services/version.test.js
```diff
@@ -93,13 +93,7 @@ describe('Versions service acceptance tests', () => {
...
         };
         await request(req);
-
-        try {
-          await request(req);
-          expect(false).toBeTruthy();
-        } catch (err) {
-          expect(err.statusCode).toEqual(400);
-        }
+        await request(req);
```
mandie722: This is a bad test, it's not actually testing anything now, you should have an expect for the positive case.

batmat: Well, it does test that the same call can be done twice without throwing an error like before, isn't it?
Care to clarify what this should 'expect'?

batmat: Done, care to have another look @mandie722 ?

PR #314, opened 2018-10-10
[JENKINS-53811] Send also current update level along with logs

batmat:
I am hoping this should help diagnosing Sentry issues, by immediately seeing which UL the instance was at when it send a given message.

Example: 
* seeing a message for an issue deemed fixed from an old UL, would orient us toward 'why this instance isn't updating?'
* seeing that message on the latest UL, where the issue triggering this message is deemed fixed already, would orient us more toward 'OK, so this is either a regression, or this is actually **not** fixed'

### DIFF
--- a/distribution/client/src/client.ts
+++ b/distribution/client/src/client.ts
@@ -43,7 +43,7 @@ export default class Client {
     this.healthChecker = new HealthChecker(process.env.JENKINS_URL || 'http://127.0.0.1:8080');
     this.update = new Update(this.app, { healthChecker: this.healthChecker });
     this.status = new Status(this.app, { flavor: process.env.FLAVOR });
-    this.errorTelemetry = new ErrorTelemetry(this.app, { flavor: process.env.FLAVOR });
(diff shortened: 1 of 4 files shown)


distribution/client/src/client.ts
```diff
@@ -43,7 +43,7 @@ export default class Client {
     this.healthChecker = new HealthChecker(process.env.JENKINS_URL || 'http://127.0.0.1:8080');
     this.update = new Update(this.app, { healthChecker: this.healthChecker });
     this.status = new Status(this.app, { flavor: process.env.FLAVOR });
-    this.errorTelemetry = new ErrorTelemetry(this.app, { flavor: process.env.FLAVOR });
+    this.errorTelemetry = new ErrorTelemetry(this.app, this.update, { flavor: process.env.FLAVOR });
```
mandie722: Why not just add this to the options map?  simplifies having to update all the contructors which is why the map is there.

batmat: I thought about this initially, but no. Does not work as contrary to the flavor, the UL changes all the time. So passing it once at the beginning in the constructor won't fly.

PR #320, opened 2018-10-11
[JENKINS-53589] delete exploded plugin dirs

mandie722:
https://issues.jenkins-ci.org/browse/JENKINS-53589

### DIFF
--- a/distribution/client/src/lib/storage.ts
+++ b/distribution/client/src/lib/storage.ts
@@ -77,6 +77,9 @@ export default class Storage {
     const pluginPath = this.pluginsDirectory();
     const retArray = [];
     plugins.forEach((plugin) => {
+      retArray.push(fs.rmdir(`${pluginPath}/${plugin}`, () => {
+        logger.info(`${pluginPath}/${plugin} was removed`);
+      }));
       retArray.push(fs.unlink(`${pluginPath}/${plugin}.hpi`, () => {
         logger.info(`${pluginPath}/${plugin}.hpi was deleted`);
         UI.publish(`Deleted ${plugin}.hpi`);
--- a/distribution/client/src/lib/storage.ts
+++ b/distribution/client/src/lib/storage.ts
@@ -1,6 +1,6 @@
 'use strict';
 
-import fs from 'fs';
+import fs from 'fs-extra';
 import path from 'path';
 
 import * as logger from 'winston';
@@ -78,9 +78,9 @@ export default class Storage {
     const retArray = [];
     plugins.forEach((plugin) => {
       retArray.push(new Promise((resolve, reject) => {
-        fs.rmdir(`${pluginPath}/${plugin}`, (err) => {
+        fs.remove(`${pluginPath}/${plugin}`, (err) => {
           if(err) {
-            reject(err);
+            logger.debug(`${pluginPath}/${plugin} was not found.`);
           }
           logger.info(`${pluginPath}/${plugin} was removed`);
           resolve(true);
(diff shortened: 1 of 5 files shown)


distribution/client/test/storage.test.ts
```diff
@@ -56,11 +56,14 @@ describe('The storage module', () => {
       const pluginPath = Storage.pluginsDirectory();
       mkdirp.sync(pluginPath);
       filenames.forEach((filename) => {
+        fs.mkdirSync(`${pluginPath}/${filename}`)
```
batmat: missing a `touchFile` in that directory to actually imitate real conditions. 

`fs.rmdir` calls http://man7.org/linux/man-pages/man2/rmdir.2.html which behaves like the `rmdir` command line, failing on non empty directories.

mandie722: I updated the test to touch a file in the directory and confirm it was all removed.

batmat: I've tested all this locally, and I strongly suspect tests are broken/ineffective for some reason. I replaced all the `.resolves` below by `.rejects` below, and still got a green build. 

However, I might still be wrong and all this is working! :)
But then as discussed on the chat, I would please request at least an explanation on why my blabbering above is wrong (about `rmdir` in Node having the same behavior as the usual command-line `rmdir`).

Thanks!

distribution/client/test/storage.test.ts
```diff
@@ -56,11 +56,17 @@ describe('The storage module', () => {
       const pluginPath = Storage.pluginsDirectory();
```
batmat: Pretty sure this is because of the `forEach`. Which could mean that nothing was really tested already here :fearful:.

I just added `expect.assertions(filenames.length*3);`  after this line, and I get 

```jest
  ● The storage module › removePlugins() › should remove all files in a list

    expect.assertions(12)

    Expected twelve assertions to be called but received zero assertion calls.

      54 |     it('should remove all files in a list', async () => {
      55 |       const filenames = ['first', 'second', 'third', 'fourth'];
    > 56 |       expect.assertions(filenames.length*3);
         |              ^
      57 |       const pluginPath = Storage.pluginsDirectory();
      58 |       mkdirp.sync(pluginPath);
      59 |       filenames.forEach((filename) => {

      at Object.it (test/storage.test.ts:56:14)
```

batmat: I believe reviewing and merging https://github.com/jenkins-infra/evergreen/pull/325 first should help.

distribution/client/src/lib/storage.ts
```diff
@@ -77,11 +77,20 @@ export default class Storage {
     const pluginPath = this.pluginsDirectory();
     const retArray = [];
     plugins.forEach((plugin) => {
+      retArray.push(new Promise((resolve, reject) => {
+        fs.remove(`${pluginPath}/${plugin}`, (err) => {
+          if(err) {
+            logger.debug(`${pluginPath}/${plugin} was not found.`);
```
batmat: I think we should analyze the error to make the log clearer, for instance:
* if ENOENT then I agree with the change and to **not** fail in case the deletion, in that case debug is probably enough
* in most (all?) other cases, then at least it should be a warning (because if there's a permission issue for whatever, we're likely to crash soon if we really need that directory to be removed)

distribution/client/src/lib/storage.ts
```diff
@@ -77,11 +77,20 @@ export default class Storage {
...
+          logger.info(`${pluginPath}/${plugin} was removed`);
+          resolve(true);
+        });
+      }));
       retArray.push(new Promise((resolve, reject) => {
 
         fs.unlink(`${pluginPath}/${plugin}.hpi`, (err) => {
           if (err) {
-            reject(err);
+            logger.debug(`${pluginPath}/${plugin}.hpi was not found.`);
```
batmat: same as above for cases where `err != ENOENT`

distribution/client/src/lib/storage.ts
mandie722: Created https://issues.jenkins-ci.org/browse/JENKINS-54207

PR #321, opened 2018-10-12
[JENKINS-54056] Update Level Automated Rollback

batmat:
[JENKINS-54056](https://issues.jenkins-ci.org/browse/JENKINS-54056)

This PR implements the automated rollback for Update Levels.
If an UL is applied to a given instance, but that instance fails to start after application, it will taint it in the backend, then ask the backend again for a previous UL (most of the time, this would be the previous one).

This PR does **not** hande the data rollback part (aka JEP-302).
This part will be handled separately in a followup as of [JENKINS-54057](https://issues.jenkins-ci.org/browse/JENKINS-54057).

### DIFF
--- a/distribution/client/src/client.ts
+++ b/distribution/client/src/client.ts
@@ -97,7 +97,9 @@ export default class Client {
       UI.publish('Jenkins appears to be online', { log: 'info' });
       Storage.removeBootingFlag();
     }).catch((error) => {
-      UI.publish('Jenkins appears to be in an unhealthy state!', { log: 'error' });
+      const errorMessage = 'Jenkins appears to be in an unhealthy state (startup auto-healthcheck)!'
+      logger.error(errorMessage);
+      UI.publish(errorMessage, { log: 'error' });
     });
 
     cron.runDaily('post-status', () => {
--- a/distribution/client/src/client.ts
+++ b/distribution/client/src/client.ts
@@ -93,13 +93,11 @@ export default class Client {
 
     this.runUpdates();
 
-    this.healthChecker.check().then((state) => {
-      if (state.healthy) {
-        UI.publish('Jenkins appears to be online', { log: 'info' });
-        Storage.removeBootingFlag();
-      } else {
-        UI.publish('Jenkins appears to be in an unhealthy state!', { log: 'error' });
-      }
+    this.healthChecker.check().then(() => {
+      UI.publish('Jenkins appears to be online', { log: 'info' });
+      Storage.removeBootingFlag();
+    }).catch((error) => {
+      UI.publish('Jenkins appears to be in an unhealthy state!', { log: 'error' });
     });
 
     cron.runDaily('post-status', () => {
--- a/distribution/client/src/client.ts
+++ b/distribution/client/src/client.ts
@@ -41,8 +41,8 @@ export default class Client {
     this.app = feathers();
     this.reg = new Registration(this.app);
     this.healthChecker = new HealthChecker(process.env.JENKINS_URL || 'http://127.0.0.1:8080');
-    this.update = new Update(this.app, { healthChecker: this.healthChecker }), { status: this.status };
     this.status = new Status(this.app, { flavor: process.env.FLAVOR });
+    this.update = new Update(this.app, { healthChecker: this.healthChecker, status: this.status });
     this.errorTelemetry = new ErrorTelemetry(this.app, this.update, { flavor: process.env.FLAVOR });
     this.updating = false;
     // This should be overridden on bootstrap
--- a/distribution/client/src/client.ts
+++ b/distribution/client/src/client.ts
@@ -97,9 +97,7 @@ export default class Client {
       UI.publish('Jenkins appears to be online', { log: 'info' });
       Storage.removeBootingFlag();
     }).catch((error) => {
-      const errorMessage = 'Jenkins appears to be in an unhealthy state (startup auto-healthcheck)!'
-      logger.error(errorMessage);
-      UI.publish(errorMessage, { log: 'error' });
+      UI.publish('Jenkins appears to be in an unhealthy state (startup auto-healthcheck)!', { log: 'error' });
     });
 
     cron.runDaily('post-status', () => {
--- a/distribution/client/src/client.ts
+++ b/distribution/client/src/client.ts
@@ -41,7 +41,7 @@ export default class Client {
     this.app = feathers();
     this.reg = new Registration(this.app);
     this.healthChecker = new HealthChecker(process.env.JENKINS_URL || 'http://127.0.0.1:8080');
-    this.update = new Update(this.app, { healthChecker: this.healthChecker });
+    this.update = new Update(this.app, { healthChecker: this.healthChecker }), { status: this.status };
     this.status = new Status(this.app, { flavor: process.env.FLAVOR });
     this.errorTelemetry = new ErrorTelemetry(this.app, this.update, { flavor: process.env.FLAVOR });
     this.updating = false;
--- a/distribution/client/src/lib/supervisord.ts
+++ b/distribution/client/src/lib/supervisord.ts
@@ -50,7 +50,7 @@ export default class Supervisord {
   static startProcess(name) {
     logger.info(`[supervisord] Starting ${name} process`);
     return new Promise((resolve, reject) => {
-      client.methodCall('supervisor.startProcess', [name], (e, value) => {
+      client.methodCall('supervisor.startProcess', [name,true], (e, value) => {
         if (e) {
           return reject(e);
         }
@@ -62,7 +62,7 @@ export default class Supervisord {
   static stopProcess(name) {
     logger.info(`[supervisord] Stopping ${name} process`);
     return new Promise((resolve, reject) => {
-      client.methodCall('supervisor.stopProcess', [name], (e, value) => {
+      client.methodCall('supervisor.stopProcess', [name,true], (e, value) => {
         if (e) {
           return reject(e);
         }
@@ -77,22 +77,6 @@ export default class Supervisord {
       await this.stopProcess(name);
     }
 
-    // During testing, it seemed like the restart call done to Supervisord could take some time
-    // for Jenkins to actually start restarting.
-    // This was causing issues, because then the healthcheck done after the restart would actually
-    // happen on the *current* version of Jenkins, running the current UL, and not the one we are
-    // about to restart "on".
-    // So we give a few seconds to actually wait for Jenkins to initiate restart so that the following
-    // healthchecks or anything else happen on Jenkins *after* restart, and not somehow during or /just before/ it.
-    return this.startProcess(name)
-      .then(() => {
-        logger.info('Waiting a few seconds while Jenkins restart is being initiated before continuing...')
-        return new Promise((resolve) => {
-          setTimeout(() => {
-            resolve(true);
-          }, 5000);
-        });
-      });
-
+    return this.startProcess(name);
   }
 }
--- a/distribution/client/src/lib/supervisord.ts
+++ b/distribution/client/src/lib/supervisord.ts
@@ -76,6 +76,23 @@ export default class Supervisord {
     if (await this.isProcessRunning(name)) {
       await this.stopProcess(name);
     }
-    return this.startProcess(name);
+
+    // During testing, it seemed like the restart call done to Supervisord could take some time
+    // for Jenkins to actually start restarting.
+    // This was causing issues, because then the healthcheck done after the restart would actually
+    // happen on the *current* version of Jenkins, running the current UL, and not the one we are
+    // about to restart "on".
+    // So we give a few seconds to actually wait for Jenkins to initiate restart so that the following
+    // healthchecks or anything else happen on Jenkins *after* restart, and not somehow during or /just before/ it.
+    return this.startProcess(name)
+      .then(() => {
+        logger.info('Waiting a few seconds while Jenkins restart is being initiated before continuing...')
+        return new Promise((resolve) => {
+          setTimeout(() => {
+            resolve(true);
+          }, 5000);
+        });
+      });
+
   }
 }
--- a/distribution/client/src/lib/update.ts
+++ b/distribution/client/src/lib/update.ts
@@ -82,8 +82,15 @@ export default class Update {
     }, {});
   }
   /*
-   * Apply the updates provided by the given Update Manifest
-   *
+   * Apply the updates provided by the given Update Manifest.
+   * Rolls back automatically to the previous non tainted level if failing.
+   * (Note: the rollback may not end up on most of the time, that would mean going from say UL-0 to U)
+
+   1) UL10 gets globally tainted after an issue is identified (UL9 is not tainted)
+   2) Instance attempts updating from UL10 to UL11, and fails
+   3) update.query() to get the right updates => receives UL9...
+   4) config files changed from UL9 to UL10, Jenkins fails to start after rolling back.
+
    * @param  {Map} Update Manifest described in JEP-307
    * @return {Promise} Which resolves once updates have been applied
    * @return {boolean} False if there is already an update in progress
@@ -137,7 +144,9 @@ export default class Update {
       this.snapshotter.snapshot(`UL${this.getCurrentLevel()}->UL${updates.meta.level} Snapshot after downloads completed, before Jenkins restart`);
       this.saveUpdateSync(updates);
       UI.publish('All downloads completed and snapshotting done, restarting Jenkins');
-      this.restartJenkins();
+    }).then( () => {
+      return this.restartJenkins();
+    }).finally( () => {
       this.updateInProgress = null;
       return true;
     });
@@ -173,7 +182,7 @@ export default class Update {
     // FIXME: actually now I'm thinking throwing in HealthChecker might provide a more
     // consistent promise usage UX here.
     // checking healthState.health value is possibly a bit convoluted (?)
-    this.healthChecker.check()
+    return this.healthChecker.check()
       .then( healthState => {
         if (healthState.healthy) {
           logger.info('Jenkins healthcheck after restart succeeded! Yey.');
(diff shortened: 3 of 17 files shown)


distribution/client/src/client.ts
```diff
@@ -93,13 +93,13 @@ export default class Client {
...
-      } else {
-        UI.publish('Jenkins appears to be in an unhealthy state!', { log: 'error' });
-      }
+    this.healthChecker.check().then(() => {
+      UI.publish('Jenkins appears to be online', { log: 'info' });
+      Storage.removeBootingFlag();
+    }).catch((error) => {
+      const errorMessage = 'Jenkins appears to be in an unhealthy state (startup auto-healthcheck)!'
+      logger.error(errorMessage);
+      UI.publish(errorMessage, { log: 'error' });
```
rtyler: Lines 101 and 102 are redundant. Passing the `{ log: 'error' }` to `UI.publish` means that it will perform a `logger.error.` underneath.

distribution/client/src/lib/supervisord.ts
```diff
@@ -76,6 +76,23 @@ export default class Supervisord {
...
+    // about to restart "on".
+    // So we give a few seconds to actually wait for Jenkins to initiate restart so that the following
+    // healthchecks or anything else happen on Jenkins *after* restart, and not somehow during or /just before/ it.
+    return this.startProcess(name)
+      .then(() => {
+        logger.info('Waiting a few seconds while Jenkins restart is being initiated before continuing...')
+        return new Promise((resolve) => {
+          setTimeout(() => {
+            resolve(true);
+          }, 5000);
```
rtyler: I'm bummed this change introduces another magic number, but I'm not sure there's any good way around this.

We can interrogate `supervisord` as to whether a process is running, perhaps `Supervisor.isProcessRunning` and t hen some timeout would be a more robust option here? Just thinking out loud.

distribution/client/src/lib/update.ts
```diff
@@ -27,6 +28,7 @@ export default class Update {
   protected readonly fileOptions : FileOptions;
   protected readonly options : any;
   protected readonly healthChecker : HealthChecker;
+  protected readonly status : Status;
```
rtyler: TYPESCRIPT WOOO

distribution/client/src/lib/update.ts
```diff
@@ -48,6 +50,12 @@ export default class Update {
     } else {
       this.healthChecker = new HealthChecker(process.env.JENKINS_URL || 'http://127.0.0.1:8080');
     }
+
+    if (this.options.status) {
+      this.status = this.options.status;
+    } else {
+      this.status = this.status = new Status(this.app, { flavor: process.env.FLAVOR });
```
rtyler: `this.status = this.status = new Status`

is the redundancy important here? Or is this just a typo?

distribution/client/src/lib/update.ts
```diff
@@ -79,24 +87,41 @@ export default class Update {
...
+    else if (this.updateInProgress || (!updates)) {
       logger.warn('applyUpdates request ignored: update already in progress!');
       return false;
+    } else {
+      // Setting this to a timestamp to make a timeout in the future
+      this.updateInProgress = new Date();
     }
+    const applyUpdatesMessage = 'Starting to apply updates';
+    logger.info(applyUpdatesMessage)
+    UI.publish(applyUpdatesMessage);
```
rtyler: This can be made into a single call, and we can remove the temporary variable with:

`UI.publish('Starting to apply updates', { log: 'info' })`

distribution/client/src/lib/update.ts
```diff
@@ -126,20 +151,30 @@ export default class Update {
     }
 
     if (tasks.length == 0) {
-      logger.debug('No actionable tasks');
+      logger.warn('No actionable tasks');
```
rtyler: I recommend adding some context to this log to make the warning more clear for debugging "No actionable tasks during upgrade process"

distribution/client/src/lib/update.ts
```diff
@@ -126,20 +151,30 @@ export default class Update {
...
     return Promise.all(tasks).then(() => {
+
       UI.publish('All downloads completed, snapshotting data before restart');
       this.snapshotter.snapshot(`UL${this.getCurrentLevel()}->UL${updates.meta.level} Snapshot after downloads completed, before Jenkins restart`);
       this.saveUpdateSync(updates);
       UI.publish('All downloads completed and snapshotting done, restarting Jenkins');
-      this.restartJenkins();
+
+    }).then( () => {
+
```
rtyler: Mind removing some of these redundant newlines?

distribution/client/src/lib/update.ts
```diff
@@ -163,57 +198,76 @@ export default class Update {
...
-  restartJenkins(rollingBack?: boolean) { // Add param to stop recursion?
-    Supervisord.restartProcess('jenkins');
+  restartJenkins() {
 
-    const messageWhileRestarting = 'Jenkins should now be online, health checking!';
+    const jenkinsIsRestarting = Supervisord.restartProcess('jenkins');
+
+    const messageWhileRestarting = 'Jenkins is being restarted, health checking!';
     UI.publish(messageWhileRestarting);
     logger.info(messageWhileRestarting);
```
rtyler: I believe I missed this in previous code reviews, similar to the prior comments, these two lines can be collapsed into one call.

distribution/client/src/lib/update.ts
```diff
@@ -163,57 +198,76 @@ export default class Update {
...
+    return jenkinsIsRestarting
+      .then(() => this.healthChecker.check())
+      .then(() => {
+        logger.info('Jenkins healthcheck after restart succeeded! Yey.');
+        return true;
+
+      }).catch((error) => { // first catch, try rolling back
+
+        const errorMessage = `Jenkins detected as unhealthy. Rolling back to previous update level (${error}).`;
+        logger.warn(errorMessage);
```
rtyler: Redundant calls :smile_cat:

distribution/client/src/lib/update.ts
```diff
@@ -163,57 +198,76 @@ export default class Update {
...
+        UI.publish(errorMessage);
+
+        return this.revertToPreviousUpdateLevel(); // FIXME: async issue!
```
rtyler: This FIXME is a little scary to me, is this the follow up work for JEP-302, or is this a bug that still needs to be filed?

distribution/client/src/lib/update.ts
```diff
@@ -163,57 +198,76 @@ export default class Update {
...
+
+      }).catch((error) => { // second time wrong, stop trying and just holler for help
+
+        // Quick notice sketch, but I do think we need a very complete and informative message
+        const failedToRollbackMessage =
+          'Ooh noes :-(. We are terribly sorry but it looks like Jenkins failed to ' +
+          'upgrade, but even after the automated rollback we were unable to bring ' +
+          'to life. Please report this issue to the Jenkins Evergreen team. ' +
```
rtyler: Please add a link to https://github.com/jenkins-infra/evergreen/issues

distribution/client/src/lib/update.ts
```diff
@@ -163,57 +198,76 @@ export default class Update {
...
+        return true;
+      });
+
   }
 
+  /*
+   * 1) Taint current level
+   * 2) query and trigger a new update
+   *    (failing UL has been marked tainted for the instance, so should be served the/a previous UL)
+   * FIXME: 'the/a previous UL' => current code will serve "a" UL. We actually want "the"
```
rtyler: This FIXME comment I am failing to parse, what's the thing to fix here?

distribution/client/test/healthchecker.test.ts
```diff
@@ -64,52 +64,66 @@ describe('The healthchecker module', () => {
...
-        expect(result.healthy).toBe(true);
-        expect(result.message).not.toContain('wrong');
+        const healthChecker = new HealthChecker(`http://localhost:${port}`, {delay: 300, retry: 1});
+        expect(await healthChecker.check()).toBeTruthy();
 
         // broken /instance-identity page
         serverOptions.instanceIdentity = false;
-        result = await healthChecker.check();
-        expect(result.healthy).toBe(false);
+        try {
```
rtyler: I believe it is better to rely on Jest for this behavior, e.g.:

```typescript
expect(async () => {
  await healthChecker.check();
}).toThrow(SomeError);
```

If this doesn't quite work properly due to the async nature of  the call, I believe adding `return` would suffice, e.g.

```typescript
return expect(async () => { }).toThrow()
```

That way Jest gets a Promise to resolve for the test.

distribution/client/test/healthchecker.test.ts
```diff
@@ -64,52 +64,66 @@ describe('The healthchecker module', () => {
...
+          // expected
+        }
 
         // issue: a plugin is failed
         serverOptions.instanceIdentity = true;
         serverOptions.metrics.plugins = false;
-        result = await healthChecker.check();
-        expect(result.healthy).toBe(false);
-        expect(result.message).toContain('wrong');
+        try {
```
rtyler: Ditto

distribution/client/test/healthchecker.test.ts
```diff
@@ -120,7 +134,7 @@ describe('The healthchecker module', () => {
   // Code below handles the very simple embedded test http server code and cases
   // See serverOptions above for behavior "configuration" of this test server
   beforeEach( async (done) => {
-    port = 3000; // Math.floor(Math.random() * (65535 - 1024) + 1024);
+    port = Math.floor(Math.random() * (65535 - 1024) + 1024);
```
rtyler: :tada:

distribution/client/src/lib/update.ts
batmat: :man_facepalming: :-), the latter

distribution/client/src/lib/update.ts
batmat: Heh, I did that on purpose as it took me so long to reflect enough to find the right sequence of calls, that I wanted to make that sequence very readable.

But yeah I guess now it can be made slightly less verbose.

distribution/client/test/healthchecker.test.ts
batmat: I tried many forms, to no easy/obvious avail for now. 

How about I address this in a followup? I would also like to take the opportunity to split that crappy (that I wrote) test in many smaller ones as much as possible.

distribution/client/src/lib/update.ts
batmat: Because of how we handle updates query, imagine the following case:

* an instance is on UL50
* for some reason, this UL is globally tainted after the fact
* a new UL51 is pushed
* the instance updates to UL51, but fails, so initiate a rollback
* it will per-instance taint UL51, then asks for where to go
* the server will tell the instance to go to UL49, not UL50. So technically, this could cause issues, especially with the data snapshoting system and the associated restore...

For now, I've ignored this issue, and granted this should almost never actually happen given host fast updates are pushed. But it's algorithmically and generally still _possible_.

So with the, granted, stupid phrasing above with _a/the_, I mean we normally expect THE previous UL (UL50 above), not just ANother UL (UL49 in the example).

Is this clearer?

batmat:
I'm pretty sure I know why the test failed above (it's unrelated). It looks to me like it's because I should add a small delay between the initial startup on UL2, and the upload of the borked UL3. Because without that delay, it can happen that the healthchecking is not fully finished yet, hence an `applyUpdates` for that new UL3 will get rejected with the usual:

```
warn: applyUpdates request ignored: update already in progress!
```

Going to add an improvement for this issue.

batmat:
BTW, I think this issue happens now (more often) because I removed a `sleep 10` too in https://github.com/jenkins-infra/evergreen/pull/321/commits/ec9f5f40223be6fe44387d815b28bc45a5d26e85#diff-62f0c255da99ad85756bb94257538d59L40 when I un-hardcoded the big `sleëp 600`.

batmat:
Confirmed, that's it.

Excerpt below: 
* the second line "_Received an Update_" is the one when the new broken UL3 is pushed.
* we can see the healthchecking finish successfully a few seconds (I figure, though we still miss https://issues.jenkins-ci.org/browse/JENKINS-53854 to be sure), but too late.
* our `wait_for_jenkins` utilitary function in tests waits for `Jenkins is fully up and running` before continuing, which is a bit wrong in that case, since we need the whole startup (healthchecking included) to have completed before pushing a new UL during test.

```
[INFO][2018-10-15 11:59:49] Jenkins is fully up and running (from hudson.WebAppMain$3 run)
info: Received an Update `created` event, checking for updates id=3, commit=container-tests-18447, [SNIP]
warn: applyUpdates request ignored: update already in progress!
info: Currently at Update Level 2
[INFO][2018-10-15 11:59:52] Obtained the latest update center data file for UpdateSource default (from hudson.model.UpdateSite updateData)
[INFO][2018-10-15 11:59:52] Obtained the updated data file for hudson.tasks.Maven.MavenInstaller (from hudson.model.DownloadService$Downloadable load)
info: Jenkins healthcheck after restart succeeded! Yey.
[INFO][2018-10-15 11:59:52] Obtained the updated data file for hudson.tools.JDKInstaller (from hudson.model.DownloadService$Downloadable load)
[INFO][2018-10-15 11:59:52] Finished Download metadata. 3,310 ms (from hudson.model.AsyncPeriodicWork$1 run)
info: Currently at Update Level 2
info: { message: 'Jenkins appears to be online',
  timestamp: 1539604793852 }
```

Anyway, fixing.

distribution/client/src/lib/supervisord.ts
```diff
@@ -50,7 +50,7 @@ export default class Supervisord {
   static startProcess(name) {
     logger.info(`[supervisord] Starting ${name} process`);
     return new Promise((resolve, reject) => {
-      client.methodCall('supervisor.startProcess', [name], (e, value) => {
+      client.methodCall('supervisor.startProcess', [name,true], (e, value) => {
```
rtyler: :+1: 

In case anybody else is looking, this is telling [supervisord to wait](http://supervisord.org/api.html#supervisor.rpcinterface.SupervisorNamespaceRPCInterface.startProcess)

distribution/client/src/lib/update.ts
```diff
@@ -79,24 +87,39 @@ export default class Update {
...
+
    * @param  {Map} Update Manifest described in JEP-307
    * @return {Promise} Which resolves once updates have been applied
    * @return {boolean} False if there is already an update in progress
    */
-  async applyUpdates(updates) {
-    if (this.updateInProgress || (!updates)) {
+  async applyUpdates(updates, forceUpdate?: boolean) {
+    if (forceUpdate) {
+      logger.warn('Forced update (expected during a rollback)');
```
rtyler: During a force update, shouldn't the `updateInProcess` still be assigned a value?

distribution/client/src/lib/update.ts
```diff
@@ -163,57 +191,70 @@ export default class Update {
...
-
-            this.snapshotter.revertToLevelBefore(this.getCurrentLevel());
-            this.revertToPreviousUpdateLevel();
-            this.restartJenkins(true);
-          }
-        }
+  restartJenkins() {
+
+    const jenkinsIsRestarting = Supervisord.restartProcess('jenkins');
+
```
rtyler: What is with all these excessive `\n\n`?

Don't bother fixing in this monolithic pull request, but kindly fix your editor, my 13" notebook only has so much vertical space :smiling_imp:

distribution/client/src/lib/update.ts
rtyler: Yes, clearer, please put this in the comments.

distribution/client/test/healthchecker.test.ts
rtyler: Fine by me to address later.

distribution/client/test/update.rollback.test.ts
```diff
@@ -0,0 +1,123 @@
...
+jest.mock('../src/lib/supervisord');
+
+describe('The update module', () => {
+  let app = null;
+  let update = null;
+
+  beforeEach(() => {
+    const evergreenHome = tmp.dirSync({ unsafeCleanup: true }).name;
+    process.env.EVERGREEN_DATA = evergreenHome;
+    Storage.homeDirectory = (() => evergreenHome);
```
rtyler: Without the `Storage` module being mocked, this seems like a pretty scary override of a module's function. I'm surprised this isn't causing test failures, but I suppose that might be because Jest is running things in parallel processes.

Perhaps it would be safer to `jest.mock` the storage module as well?

I see the same code in `update.test.js`, which means I missed this in a previous set of patches, so it can be addressed in a follow up, but I think it should be addressed.

distribution/client/src/lib/supervisord.ts
batmat: FTR, associated dedicated commit for this is https://github.com/jenkins-infra/evergreen/pull/321/commits/79a5eb02d933e023b737f6ecae671b21e5c28170

distribution/client/src/lib/update.ts
batmat: Good point. Normally, though it's indeed not enforced, the current code is expecting that calling force is used when calling itself recursively for rolling back, so `updateInProcess` should already have a value. 

Going to have a look, probably will mainly add a warning if not.

distribution/client/src/lib/update.ts
batmat: It's manually hard-wrapped for readability. So nothing I can fix on the editor side :-).

More seriously, I could _probably_ remove one or two newlines, but looking at the whole `restartJenkins` method, it does not look like there's that much unnecessary newlines there.
I guess I can remove L197 and L209, but the rest looks mostly fine to me.

image

ah #opinions :-)

distribution/client/src/lib/update.ts
batmat: done

distribution/client/src/lib/update.ts
batmat: fixed with last push

batmat:
Going to move forward and merge this. But obviously, I thoroughly welcome an additional review @mandie722, I'll file followup PRs in case you find something here that needs fixing.

Thanks!

PR #323, opened 2018-10-17
Update package-lock.json files and add tasks for it on top level Makefile

batmat:
TBH I agree it looks a bit scary, but at the same time, it seems to we should be doing this regularly.
Or should we only run `npm audit fix` on a regular basis?

Not sure what people do in general wrt. updating the `package-lock.json` file. Reading the Internet, seems like people generally say this should, like I did here, just be deleted then `npm install` be run again. But given the amount of changes this generates, this always looks a bit dangerous.

I guess the right/intermediate way is also to do that more regularly to reduce the size of the diff. CD FTW.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,3 +17,13 @@ fix-formatting:
 publish:
 	$(MAKE) -C distribution $@
 	$(MAKE) -C services $@
+
+update-package-locks:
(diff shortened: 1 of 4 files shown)


batmat:
> This is a lot of upgrades, I hope you've done some manual testing as well

yup :)

PR #324, opened 2018-10-17
[WIP] Enable ESLint typescript parsing

batmat:


Makefile
```diff
@@ -17,3 +17,13 @@ fix-formatting:
 publish:
 	$(MAKE) -C distribution $@
 	$(MAKE) -C services $@
+
+update-package-locks:
+	# Sigh, the sed for forcing https below is recommended by the npm registry team itself...
```
rtyler: gross!

distribution/client/src/lib/update.ts
```diff
@@ -22,6 +22,8 @@ export interface FileOptions {
   flag?: string,
 };
 
+interface Foo {}
```
rtyler: I assume that this is here just to fail the build on purpose?

batmat: yes, but does not work, hence the WIP state :cry:

batmat:
Closing for now.
Anyway, I'm leaning towards trying TSLint instead since it makes more sense to go the typescript direction, instead of trying to refuse ESLint for Typescript.

PR #331, opened 2018-10-23
[JENKINS-54204] Make sure to use the correct dependencies

alecharp:
It can happen with `npm install` that we have new / different dependencies installed, which could be seen as `package-lock.json` is modified. Using `npm ci` rather than `npm install` make sure this doesn't happen.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,9 @@ update-package-locks:
 	# https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/13
 	rm -rf services/node_modules/ services/package-lock.json && \
 	rm -rf distribution/client/node_modules/ distribution/client/package-lock.json && \
-	cd services/ &&	$(NODE) npm install && \
+	cd services/ &&	$(NODE) npm ci && \
 	sed -i 's/"resolved": "http:/"resolved": "https:/g' package-lock.json && \
-	cd ../distribution/client && $(NODE) npm install && \
+	cd ../distribution/client && $(NODE) npm ci && \
 	sed -i 's/"resolved": "http:/"resolved": "https:/g' package-lock.json
 
 npm-audit-fix:
--- a/distribution/Makefile
+++ b/distribution/Makefile
@@ -46,7 +46,7 @@ prepare-distribution: build/jenkins-support build/jenkins.sh
 
 	# Once the tree has been put into place, node_modules must be generated for
 	# the distribution
-	(cd $(DIST_DIR)/client && $(ROOT_DIR)/tools/npm install --production)
+	(cd $(DIST_DIR)/client && $(ROOT_DIR)/tools/npm ci --production)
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -23,9 +23,9 @@ update-package-locks:
 	# https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/13
 	rm -rf services/node_modules/ services/package-lock.json && \
 	rm -rf distribution/client/node_modules/ distribution/client/package-lock.json && \
-	cd services/ &&	$(NODE) npm install && \
+	cd services/ &&	$(NODE) npm ci && \
```
batmat: I think :bug:.
This one is actually designed to touch/update the `package-lock.json` files. So, AIU the `npm ci` task, I suspect this would just fail because there's *no* package-lock file when it's called.

Going to check locally.

distribution/Makefile
```diff
@@ -46,7 +46,7 @@ prepare-distribution: build/jenkins-support build/jenkins.sh
 
 	# Once the tree has been put into place, node_modules must be generated for
 	# the distribution
-	(cd $(DIST_DIR)/client && $(ROOT_DIR)/tools/npm install --production)
+	(cd $(DIST_DIR)/client && $(ROOT_DIR)/tools/npm ci --production)
```
batmat: That one looks :+1:

services/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM node:9 as builder
+FROM node:10 as builder
```
batmat: Not sure how this is related, this looks like it's not, and it should have been a separate PR. (?)

Makefile
batmat: Confirmed

```
$ make update-package-locks                                                                                                                                                                                                       d04b1ce
# Sigh, the sed for forcing https below is recommended by the npm registry team itself...
# https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/13
rm -rf services/node_modules/ services/package-lock.json && \
rm -rf distribution/client/node_modules/ distribution/client/package-lock.json && \
cd services/ &&  npm ci && \
sed -i 's/"resolved": "http:/"resolved": "https:/g' package-lock.json && \
cd ../distribution/client &&  npm ci && \
sed -i 's/"resolved": "http:/"resolved": "https:/g' package-lock.json
npm ERR! cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json with lockfileVersion >= 1. Run an install with npm@5 or later to generate it, then try again.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/tiste/.npm/_logs/2018-10-25T09_11_12_808Z-debug.log
make: *** [Makefile:24: update-package-locks] Error 1
```

batmat: Fixed it myself with last commit.

PR #346, opened 2018-11-08
[JENKINS-54466] Plugin extension should be jpi instead of hpi

batmat:
[JENKINS-54466](https://issues.jenkins-ci.org/browse/JENKINS-54466)

### DIFF
--- a/distribution/client/src/lib/storage.ts
+++ b/distribution/client/src/lib/storage.ts
@@ -1,10 +1,15 @@
 'use strict';
 
-import fs from 'fs-extra';
+import fs   from 'fs-extra';
 import path from 'path';
 
 import * as logger from 'winston';
 import UI from './ui';
+
+import util from 'util';
+
+const remove = util.promisify(fs.remove);
+const unlink = util.promisify(fs.unlink);
 
 /*
  * The Storage module simply contains common functions necessary for the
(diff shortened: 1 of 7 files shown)


distribution/client/src/lib/storage.ts
```diff
@@ -98,6 +103,18 @@ export default class Storage {
...
+        fs.unlink(`${pluginPath}/${plugin}.jpi`, (err) => {
+          if (err) {
+            logger.debug(`${pluginPath}/${plugin}.jpi was not found.`);
+          }
+          logger.info(`${pluginPath}/${plugin}.jpi was deleted`);
+          UI.publish(`Deleted ${plugin}.jpi`);
+          resolve(true);
+        });
+
+      }));
```
batmat: Going to file a followup dedicated PR to promisify it. I didn't want to mix feature addition, and reworking all this (because there are three similar blocks)
https://issues.jenkins-ci.org/browse/JENKINS-54545 to track this.

distribution/client/src/lib/storage.ts
```diff
@@ -98,6 +103,18 @@ export default class Storage {
         });
 
       }));
+      retArray.push(new Promise((resolve, reject) => {
+
+        fs.unlink(`${pluginPath}/${plugin}.jpi`, (err) => {
+          if (err) {
+            logger.debug(`${pluginPath}/${plugin}.jpi was not found.`);
```
jvz: Do you want to reject here?

batmat: @jvz not *really*. We should actually reject or at least in case it's something else than a file not found, but it's a known issue we are going to tackle very soon https://issues.jenkins-ci.org/browse/JENKINS-54207

rtyler: Since there's not a `stat` operation happening before this, an error might be non-fatal (i.e. "not found". I don't think rejecting would make sense on this tbh

PR #347, opened 2018-11-08
npm audit fix

batmat:
It seems like `npm audit fix` adds optional: true for `devDependencies`. 

I'm not sure where it comes from, but it seems consistent over the few executions I've done in the last few weeks. So, my hope is that this changes once and doesn't come back and forth. If so, we will see what we want to do with this...

### DIFF
--- a/distribution/client/package-lock.json
+++ b/distribution/client/package-lock.json
@@ -3718,7 +3718,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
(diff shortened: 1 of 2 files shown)


distribution/client/package-lock.json
```diff
@@ -6587,9 +6592,9 @@
       }
     },
     "merge": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/merge/-/merge-1.2.0.tgz",
-      "integrity": "sha1-dTHjnUlJwoGma4xabgJl6LBYlNo=",
+      "version": "1.2.1",
```
batmat: This is the main update.

PR #348, opened 2018-11-11
Update of plugins already declared under spec

batmat:
Left the ones only in status out of this for now, since it means adding them to the spec already.

What's left to be done after this:

```
$ node ./prepare-essentials propose-updates --uc ./update-center.json
info: The update center has a newer branch-api: 2.0.21
info: The update center has a newer cloudbees-bitbucket-branch-source: 2.2.14
info: The update center has a newer durable-task: 1.27
warn: No such plugin essentials
warn: No such plugin evergreen
info: The update center has a newer scm-api: 2.3.0
info: The update center has a newer workflow-basic-steps: 2.12
info: The update center has a newer workflow-durable-task-step: 2.26
```


jvz (approved):

services/essentials.yaml
```diff
@@ -62,7 +62,7 @@ spec:
       version: '1.17'
     - groupId: org.jenkins-ci.plugins
       artifactId: jackson2-api
-      version: 2.9.7.0-beta
+      version: 2.9.7.1
```
alecharp: others are in `'`, but not this one. expected?

services/essentials.yaml
```diff
@@ -89,16 +89,16 @@ spec:
...
       artifactId: workflow-scm-step
       version: '2.7'
     - groupId: org.jenkins-ci.plugins.workflow
       artifactId: workflow-support
-      version: '2.21'
+      version: '2.22'
     - groupId: org.jenkins-ci.plugins
       artifactId: pipeline-utility-steps
-      version: 2.1.0
+      version: 2.2.0
```
alecharp: same question about quotes

alecharp (approved):
Just one question: some versions are in quotes (`'`) but not all. Is there a reason for this?

services/essentials.yaml
batmat: shitty version parser and JSON FTW. Will be parsed as numeric or string depending mostly randomly :). Also it tries generally to compare using semver rules :man_facepalming:.
TL;DR: We need to replace that component, but it's not critical for now.

mandie722 (approved):

services/essentials.yaml
jvz: YAML will parse floats as floats basically. So version `1.10` would be parsed as `1.1`, whereas version `'1.10'` will be parsed as the string `1.10`. Versions with at least two dots won't be parsed as numbers, though, hence the shorthand here.

batmat: Yeah, I tried it. But basically, once you add things under `spec`, then using `make -C services generate-essentials` will compute what's needed under `status` and rewrite the whole file removing some of these. Again I didn't take time to dig into that either.

jvz: YAML parsers will usually print out the YAML in the simplest way it can, so it'll only quote version strings that are ambiguous. There's probably a formatting option in whichever YAML library you're using here.

PR #350, opened 2018-11-13
Revert to JCasC 1.1 to avoid dozens of warnings for core classes

batmat:
Temporary while at least the core warnings are being fixed.
We should probably add tests to check the number of warnings generated in the json file parsed by the evergreen-client and sent to the Error Telemetry backend.

Manually tested on my own instance that going back to a previous version of CasC my Pipeline still runs successfully like before.

Rolling back to the state before https://github.com/jenkinsci/configuration-as-code-plugin/pull/587 while core warnings are being fixed.

### DIFF
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -8,7 +8,7 @@ spec:
   plugins:
     - groupId: io.jenkins
       artifactId: configuration-as-code
-      version: '1.3'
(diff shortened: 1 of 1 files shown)


jvz (approved):

rtyler (approved):
:man_facepalming:

PR #351, opened 2018-11-14
[JENKINS-54598] Create Java 11 build structure

alecharp:
This is to address [JENKINS-54598](https://issues.jenkins-ci.org/browse/JENKINS-54598)

- [x] create Java 11 flavor
- [ ] add plugins compatible with Java 11
- [x] validate java version in image
- [ ] validate java 11 image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -164,6 +164,17 @@ pipeline {
                       }
                   }
                 }
+                stage('Java 11 Docker Cloud image') {
+                  agent { label 'linux' }
+                  steps {
+                      sh 'make -C distribution clean java11-docker-cloud-container-check'
+                  }
+                  post {
+                      always {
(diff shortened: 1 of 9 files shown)


batmat:
For people watching, we've been working a bit on this together off the band with Adrien. The bottom line is that because of Alpine not having (yet?) a JDK11 available, this is proving very hard to move forward.

@alecharp Can you please see possibly to dump a brain dump of the current state and your experiments as a comment in https://issues.jenkins-ci.org/browse/JENKINS-54598 ?

Thanks a lot!

batmat:
I believe the two commits added on top of this in #353 should do the trick. :crossed_fingers:

alecharp:
@batmat Could you put those commits on top of my branch? As you have push access to the origin repository, you have push access to repositories with branches which are PR in the original repository.

PR #355, opened 2018-11-22
Raise build timeout for better trusted.ci accomodation

batmat:
Trusted.ci Jenkins infra behind the VPN can sometimes have many agents failed but still counting as used.
Given this infra is less visible and monitored than ci.jenkins.io, this can go unnoticed for quite some time.

In the current case, it could make Evergreen images publishing fail there because we reach the 1 hour timeout because all parallel branches are then run serially instead of in parallel (the build takes normally ~30 min on both ci.jenkins.io and trusted.ci as we could see a minute ago on a call with @olblak).

So I'm bumping this global build timeout so we're less sensitive to this potential infra issues.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@ pipeline {
     agent { label 'linux' }
 
     options {
-        timeout(time: 1, unit: 'HOURS')
+        timeout(time: 4, unit: 'HOURS')
         buildDiscarder(logRotator(daysToKeepStr: '10'))
         timestamps()
     }


alecharp (approved):
Could we have a "retry" policy for such cases? But having a bigger timeout is fine for me too.

batmat:
Eeeek 
https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.1%2B13/OpenJDK11-jdk_x64_linux_openj9_11.0.1_13.tar.gz leads to a 404...

batmat:
> Could we have a "retry" policy for such cases? But having a bigger timeout is fine for me too.

We cannot really. The failure on trusted.ci **is** due to the timeout of 1 hour when the infra has not enough agents available. So retrying inside that global timeout rule is not going to work. And I do not think we can configure a global retry for this if that's what you mean.

batmat:
Failure cause in review in #356, will reopen here once the other is merged.

PR #362, opened 2018-11-29
Add start-dev.sh script for ease of first time setup

jennyfive:
This is a temporary solution to prevent the error `error: Error during call: not-modified: No updates presently available` that pops by if the `curl` call is not done as fast as the `backend server` is up. Open for suggestions!


jennyfive:
:sparkles: !

PR #366, opened 2018-11-30
JENKINS-54545 Promisify plugin deletion

jenni:
https://issues.jenkins-ci.org/browse/JENKINS-54545

Giving a try to promisifying the plugin deletion in the `storage.ts` file. Not sure whether it's correct to handle errors on each promise or in `Promise.all` instead.

### DIFF
--- a/distribution/client/src/lib/storage.ts
+++ b/distribution/client/src/lib/storage.ts
@@ -82,39 +82,29 @@ export default class Storage {
     const pluginPath = this.pluginsDirectory();
     const retArray = [];
     plugins.forEach((plugin) => {
-      retArray.push(new Promise((resolve, reject) => {
-        fs.remove(`${pluginPath}/${plugin}`, (err) => {
-          if(err) {
-            logger.debug(`${pluginPath}/${plugin} was not found.`);
-          }
-          logger.info(`${pluginPath}/${plugin} was removed`);
-          resolve(true);
-        });
-      }));
-      retArray.push(new Promise((resolve, reject) => {
-
-        fs.unlink(`${pluginPath}/${plugin}.hpi`, (err) => {
-          if (err) {
-            logger.debug(`${pluginPath}/${plugin}.hpi was not found.`);
-          }
-          logger.info(`${pluginPath}/${plugin}.hpi was deleted`);
-          UI.publish(`Deleted ${plugin}.hpi`);
-          resolve(true);
-        });
-
-      }));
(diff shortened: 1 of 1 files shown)


mandie722 (commented):
There is a problem with removing the error checks which ties in to https://issues.jenkins-ci.org/browse/JENKINS-54207.  There does need to be a path for throwing an error in the event that the file is not accessible (vs not found which is not a bad error)

rtyler (approved):

distribution/client/src/lib/storage.ts
```diff
@@ -82,39 +82,29 @@ export default class Storage {
...
-          resolve(true);
-        });
-
-      }));
+      retArray.push(
+        remove(`${pluginPath}/${plugin}`)
+          .then(() => {
+            logger.info(`${pluginPath}/${plugin} was deleted.`);
+            return true;
+          }).catch(() => logger.debug(`${pluginPath}/${plugin} was not found.`))
```
batmat: This LGTM: we will to reintroduce the `error` parameter when we implement https://issues.jenkins-ci.org/browse/JENKINS-54207. AFAIU, this change does keep the current behavior as-is.

The current behavior is actually wrong, but the good news is that it be your next friend @jennyfive now you're on a roll ;-): if you want to give a try to https://issues.jenkins-ci.org/browse/JENKINS-54207, we'll be happy to help :)

batmat (approved):

batmat:
I checked: only the client side was touched here, so the CI failure is unrelated.

image

batmat:
Wuut, now the client is failing on something that could seem related.

image

image

jennyfive:
@_@ the tests were passing locally, it's a bit difficult to keep up now here. Could it be that there were some new merged changes that affect this PR? I can rebase and look into it again, any tip on how-why-what's going on?

batmat:
@jennyfive in case you were wondering, I didn't intend to keep this closed. I was planning to reopen it to retrigger CI but forgot :man_facepalming:.

TBH, I do not think your change is the "culprit" here... Some kind of flakiness I think.

batmat:
:fireworks: thanks @jennyfive!!! 

So next stop [JENKINS-54207](https://issues.jenkins-ci.org/browse/JENKINS-54207) ? :yum:

jennyfive:
@batmat happily! :rocket:
Thanks @mandie722 for the hint up there about the `e`.

PR #372, opened 2019-01-07
Upgrade to the latest available plugins in the Pipeline suite

rtyler:
While I'm still not sure what matrix of these plugins are _supposed_ to work
properly together. The latest from the Update Center might be okay :hankey:

:fire: **note**: this upgrades the `docker-cloud` flavor's version of `workflow-support` from an incremental build to the latest 3.0 :fire:

### DIFF
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -187,9 +187,6 @@ spec:
         - groupId: io.jenkins.docker
           artifactId: docker-plugin
           version: 1.1.6-rc865.abf142391212
-        - groupId: org.jenkins-ci.plugins.workflow
-          artifactId: workflow-support
-          version: 2.23-rc674.73bb98187744
 status:
   core:
     version: '2.156'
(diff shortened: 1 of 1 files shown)


rtyler:
Because of the change for the `docker-cloud` plugin, this should not be merged without @batmat's approval

services/essentials.yaml
```diff
@@ -78,27 +78,45 @@ spec:
...
     - groupId: org.jenkins-ci.plugins.workflow
       artifactId: workflow-scm-step
       version: '2.7'
+    - groupId: org.jenkins-ci.plugins.workflow
+      artifactId: workflow-step-api
+      version: '2.17'
     - groupId: org.jenkins-ci.plugins.workflow
       artifactId: workflow-support
-      version: '2.22'
+      version: '3.0'
```
batmat: That's fine. That's a great fix we were waiting on for the Java 11 support, and in the end is also compatible with Java 8.
And we don't expect issues with it.

batmat:

PR #373, opened 2019-01-08
Update to plugins fixed in security advisory 2019-01-08

batmat:
https://jenkins.io/security/advisory/2019-01-08/


alecharp (approved):
I guess the releases are still private, because I couldn't see the tags for those releases on any repository at the time I review this PR. But you must know the version to use so :+1:

PR #376, opened 2019-01-22
Update to Jenkins 2.161

batmat:


alecharp (approved):
this is something we might endup doing every week..

batmat:
@alecharp yes. This is something we must do constantly. I have started a task to automate it some months ago, but I got stalled before finalizing it. It was partly working.

PR #378, opened 2019-01-22
[JENKINS-55547] Switch to Centos as base image

batmat:
[JENKINS-55547](https://issues.jenkins-ci.org/browse/JENKINS-55547) Main goal is to have an easier path for future JDK upgrades, which is very risky for Alpine.

Cf. https://twitter.com/mreinhold/status/1046862466204950529

### DIFF
--- a/distribution/Dockerfile
+++ b/distribution/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10-alpine
+FROM centos:7.6.1810
 
 ARG user=jenkins
 ARG group=jenkins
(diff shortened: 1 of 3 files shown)


alecharp (approved):
LGTM

rtyler (approved):

batmat:
@jenkins-infra/java11-support FYI

distribution/Dockerfile
```diff
@@ -50,18 +50,33 @@ EXPOSE ${agent_port}
...
 
+# Use the auto-install script, but prevent any unnoticed change
+# https://linuxize.com/post/how-to-install-node-js-on-centos-7/
+RUN curl -sL https://rpm.nodesource.com/setup_10.x > /tmp/install-node && \
+    echo "2050d9584f54fd1da7dfc54f52eac24d  /tmp/install-node" | md5sum --check && \
+    cat /tmp/install-node| bash - && \
+    rm /tmp/install-node
+RUN yum install -y --setopt=skip_missing_names_on_install=False \
+    nodejs
+RUN node --version | grep '^v10.' > /dev/null && \
```
batmat: I will spend a bit of time later to optimize layers, I could probably save some disk space and download time here.

batmat: Also about the `grep`, there are obviously not needed in terms of image, but it saved a lot of time, and I feel is good to catch it early in the future.

We have tests in `offline-tests.sh` that check that too, so in general these tests are _almost_ moot now. Though I feel like they are still interesting for when/if we create new flavors and we don't handle node/npm versions in new ``Dockerfile``s

distribution/scripts/start-client.sh
```diff
@@ -24,4 +24,5 @@ else
   echo "Client is starting up"
 fi
 
-exec /usr/local/bin/npm run start
+export PATH=/usr/bin:/usr/local/bin:$PATH
```
batmat: FTR: I did this this way to be both compatible with the new and old PATH. (former is the new PATH for `npm` on Centos, the latter was the one on `alpine`).

distribution/tests/offline-tests.sh
```diff
@@ -14,7 +14,7 @@ oneTimeSetUp() {
 
 # JENKINS-49864
 test_docker_CLI_available() {
-  docker exec "$container_under_test" which docker > /dev/null
+  docker exec "$container_under_test" docker --version > /dev/null
```
batmat: Believe it or not, Centos does not have `which` installed by default.

PR #380, opened 2019-01-28
Update various plugins

batmat:


services/essentials.yaml
```diff
@@ -35,7 +35,7 @@ spec:
       version: '1.17'
     - groupId: org.jenkins-ci.plugins
       artifactId: git
-      version: 4.0.0-rc2892.70634f157021
+      version: 4.0.0-rc3016.df016c238c06
```
batmat: @markewaite does this look right to you? Thanks!

MarkEWaite: That is an incremental build from today.  I've checked that it exists in artifactory.

Within the next 1-2 days I plan to deliver the 4.0.0 release candidate build.  It will stay in "release candidate" mode for at least a month.  It will be based on the master branch, same as the incremental that you're referencing here.

MarkEWaite (approved):
Confirmed the git plugin and git client plugin both exist at the specified version and confirmed that the modified plugins match the same changes I made to a training image definition for the latest security update.

rtyler (approved):

PR #386, opened 2019-02-12
Update to latest 2.164

batmat:
To-do in general, but even more important these days so we can switch to a JDK11 in some not too distant future.

### DIFF
--- a/services/essentials.yaml
+++ b/services/essentials.yaml
@@ -4,7 +4,7 @@ metadata:
   annotations: null
 spec:
   core:
-    version: '2.163'
+    version: '2.164'
   plugins:
     - groupId: io.jenkins
       artifactId: configuration-as-code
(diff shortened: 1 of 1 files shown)


rtyler:
The failures in ci.j.io _appear_ to be legitimate, but I haven't had the opportunity to dig deeper into the logs.

batmat:
@rtyler yeah, had a look earlier, but no time to dig. I agree this looks like an actual issue.

Going to retrigger to have an additional hint during my night, but I think something in the `jenkins.sh` file we download and use as is was changed recently.

I strongly suspect also our mistake is to use the `master` branch instead of a fixed sha in https://github.com/jenkins-infra/evergreen/blob/master/distribution/Makefile#L7

batmat:
Filed also https://github.com/jenkins-infra/evergreen/pull/387 to see if this fails too on current master without any actual change.

alecharp (approved):

oleg-nenashev (approved):

PR #388, opened 2019-02-14
Fix java11 flavor current test failure

batmat:
The upstream jenkins-jdk11.sh file has been removed.
Given now the base image is CentOS based, that commit removes the full
Dockerfile override and changes it to just override the JDK to 11.

I also pinned the version of files we use from https://github.com/jenkinsci/docker to avoid this in the future to surface in some seemingly random manner.


batmat:
Just merged with https://github.com/jenkins-infra/evergreen/pull/386, to allow running Jenkins without `--enable-future-java` flag.

alecharp (approved):
LGTM

PR #392, opened 2019-03-01
Upgrade to Configuration as Code 1.7

rtyler:


batmat:
@rtyler conflicted with just merged #393 I believe :\

rtyler:
Out of order merges will do that

PR #395, opened 2019-03-06
Update script-security to 1.54

batmat:


rtyler (approved):
yay more script-security updates :man_facepalming:

PR #403, opened 2019-04-08
Update a swath of plugins

rtyler:


batmat:
Probably a client build flakiness. Had been long since I saw this last. Retrying.

PR #404, opened 2019-04-10
Update to Jenkins 2.172

batmat:


rtyler (approved):

batmat:
Failure is real, and due to the removal of the remoting-based CLI in https://github.com/jenkinsci/jenkins/pull/3838, released in 2.165 weekly.

We _could_ easily fix the build by removing the configuration in the Evergreen CasC file in https://github.com/jenkins-infra/evergreen/blob/9c9e2141ef2e7d7321abd079e2056d190a6f7ba7/distribution/config/as-code/create-admin-user.yaml#L20-L21, but this would then definitely crash Evergreen instances out there, because we didn'd implement yet the update of the config files.

I'm thinking I could probably use the Evergreen Jenkins plugin to update these files. This is likely a bit hackish/dirty, but I think given this is a security update, and we need to move forward, this could be the easiest way out...

WDYT @rtyler ?

batmat:
Probably a short term fix could be to go to 2.164.2, but then it would just postpone the need to find a resolution for this issue...

rtyler:
Hah, I bet this failure will also affect some of my non-Evergreen instances too. Hm. I'll need to ponder a bit here, this is sticky

batmat:
I wonder if we could (should?) reintroduce some kind of empty-shell object inside Jenkins matching this `remotingCLI` name, but which would do nothing for some but spit big far warnings, telling them for instance "**please fix your config file, this `remotingCLI` name thingy will be wiped out on 2020-04-01, lol srsly roff**"?

This way, users using JCasC and this field would have some time to fix their config-as-code without everything blowing up, but the attribute would just become ignored?

WDYT @jglick does this sound like an idea/feasible?

----

Side note: this wouldn't solve the issue on Evergreen side, but at least also this would also give us some more time to figure out the right way to handle this kind of issue.

jglick:
If I remember @ndeloof’s recommendations correctly, it was that a given JCasC YAML file was valid for a particular version of Jenkins core and plugins, and may or may not be valid after updates. Thus if Evergreen manages these YAML files and manages component updates, it is on hook to edit the YAML alongside the update.

batmat:
See #409 for an updated version.

PR #406, opened 2019-04-22
Remove the disappeared remotingCLI configuration

rtyler:
In the latest weeklies, this configuration disappeared entirely and
Configuartion-as-Code users will experience hard-crashes on boot with this
configuration present.

The changes to how Configuration as Code is configured in this pull request does mean that we wouldn't see hard failures in CI for bad configuration @batmat , but I think it's worth the _potential_ to miss configuration errors.

### DIFF
--- a/distribution/config/as-code/casc.yaml
+++ b/distribution/config/as-code/casc.yaml
@@ -1,0 +1,8 @@
+---
+# Avoid hard errors on things which o not exist
+configuration-as-code:
+  version: 1
+  deprecated: warn
+  restricted: warn
+  unknown: warn
+
--- a/distribution/config/as-code/create-admin-user.yaml
+++ b/distribution/config/as-code/create-admin-user.yaml
@@ -16,6 +16,3 @@ jenkins:
   crumbIssuer:
     standard:
       excludeClientIPFromCrumb: false
-security:
-  remotingCLI:
-    enabled: false


batmat:
I agree overall, even more because we don't have the feature yet to update the evergreen-client itself. 
Ignoring errors is bad.

However: as a double-check @rtyler, so this change is **actually** useful, did you check that:
* adding only `distribution/config/as-code/casc.yaml`, 
* and bumping to latest weekly, 
* **and** NOT removing the `remotingCLI` config should make the PR go green

batmat (changes requested):
to clarify, let's not land this before the fix in flight in #408 so we actually have still the entry to check in there.

batmat:
After a second check, I'd actually be more confident if this PR was two commits:
1. First with just `distribution/config/as-code/casc.yaml` to show the failure doesn't show up anymore (we need to revert #408 somehow temporarily in this commit so this actually checks errors are now warnings only)
2. Remove the remotingCLI config

rtyler:
@batmat considering the changes we have since merged which provide this symbol, I don't think the previous feedback applies

rtyler:
@batmat heh, why'd you close my pull request? :laughing: 

I still think the change should be applied ,I just don't think [this](https://github.com/jenkins-infra/evergreen/pull/406#issuecomment-491629658) is possible

batmat:
:man_facepalming: forgot to reopen to retrigger a build. Pretty sure I had even a comment ready and didn't click "Reopen and comment". Eeeeeeeeek

PR #407, opened 2019-04-22
Do not start multiple builds in parallel

daniel-beck:
If there's something wrong with executors, this just keeps adding new builds to the queue otherwise.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@ pipeline {
         timeout(time: 4, unit: 'HOURS')
         buildDiscarder(logRotator(daysToKeepStr: '10'))
         timestamps()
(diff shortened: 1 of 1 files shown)


batmat:
So, I thought this was fine, but finally it's a bit annoying. That's OK here given the level of activity of Evergreen, but I very often use multiple commits to test something is failing, or fixed. Like right now in https://github.com/jenkins-infra/evergreen/pull/408, and with this change, the second commit will only be assessed after the first has finished... And given the duration of the full build, this build time serialization really adds up.

Anyway again, that is fine at least for Evergreen as it is now I think.
We'll see how this flies in the long run.

daniel-beck:
You can exclude PRs from this (or at least you can in scripted)

PR #408, opened 2019-04-26
Update to latest weekly / handle remotingCLI symbol removal

batmat:
Shows https://github.com/jenkinsci/evergreen-plugin/pull/20 in use

### DIFF
--- a/distribution/tests/tests.sh
+++ b/distribution/tests/tests.sh
@@ -290,7 +290,7 @@ test_no_hpi() {
 test_no_remotingCLI_in_casc_file() {
 
   # race condition risk? That check might run *before* the plugin has had the time to modify the file?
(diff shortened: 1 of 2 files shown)


batmat:
First commit is just a starting point, showing that 2.165 will fail if done alone.

batmat:
Will cleanup the PR on Evergreen-plugin before merge, the code is not clean and documented enough, but the concept was proved to work.

rtyler:
@batmat the tests are all passing, is this still a WIP?

batmat:
@rtyler given you fixed the test afaiu, let's go ahead.

However I'm very AFK. I think it would be good to spawn a new instance now **before** merging to check the update actually works fine. WDYT ?

batmat:
Have an instance running locally the current UL. 
image

Will check if it updates fine to upcoming UL190

batmat:
yey 
image

PR #426, opened 2019-08-03
Update the node setup script for with its new checksum

rtyler:
Checked the latest and it looks legit


rtyler:
abusing super powers!

PR #437, opened 2019-09-09
[JENKINS-59287] Fix README instructions for IPv6

candrews:


jsoref:
I think I'd like the commit message to explain that `ident.me` may return an ipv4 or ipv6 address but that this command expects an ipv4 address so the change is to provide the input it's expecting.

With that, this seems like a perfectly reasonable change.

candrews:
> I think I'd like the commit message to explain that `ident.me` may return an ipv4 or ipv6 address but that this command expects an ipv4 address so the change is to provide the input it's expecting.

Sounds good to me - I've updated this PR accordingly.

jsoref:
> However, this command expects an ipv4 address so this change provides the input (and ipv4 address) it's expecting.

Definitely better, do you mean `and ipv4` or `an ipv4`?

candrews:
> Definitely better, do you mean `and ipv4` or `an ipv4`?

:man_facepalming: 

Updated the PR.

jsoref:
Oh, I'd probably change the PR summary, something like:

> Fix README instructions to not fail if user has an IPv6 address

(It took me a couple of readings to understand why one would be providing an IPv4 address as the change for a PR that was talking about IPv6)

Sorry for the small input, and thanks for working on IPv6 support.

candrews:

PR #438, opened 2019-09-09
Update CloudFormation to use the latest t3 and m5 instance types

### LINKED ISSUE #432: CloudFormation: Permit newer instance generations
Allow users to select the new t3 instances instead of t2.
New t3 instances requires ENA enabled AMIs, and the one you are currently using on the template should be ok.

candrews:
Add support for t3 and m5 instance types
Change default from t2.large to t3.large

Closes #432


candrews:

PR #439, opened 2019-09-09
Parameterize the AMI and use the latest by default

### LINKED ISSUE #431: CloudFormation: Add support for other regions
If you try to launch this Stack in another region than Virginia, it will fail because it won't be able to locate the specified AMI ID.

In order to support other regions, a region-ami_id mapping would need to be added into the template.

candrews:
Adds a parameter for the AMI to use for the Evergreen instance
Use the latest AMI by default
By using a system parameter store parameter, the reference works across all regions which closes #431


candrews:
@jsoref can you please review this and hopefully merge it?

candrews:

PR #441, opened 2019-09-10
Specify the region to use in the Jenkins configuration

### LINKED ISSUE #431: CloudFormation: Add support for other regions

candrews:
Previously was hardcoded to us-east-1 so evergreen could only work in that region; this PR fixes that issue.

Closes #431


candrews:
@jsoref can you please review this and hopefully merge it?

PR #442, opened 2019-09-10
Recover the Evergreen instance if its instance health check fails for 5 minutes

candrews:
See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-recover.html


candrews:

